PR #8, opened 2020-01-05
fix systemd unit file install path

aroundthfur:
some distros do not have the `lib -> usr/lib` symlink


srbhp:
Thanks.

AdnanHodzic:
@srbhp thanks for reporting and @aroundthfur for fixing this issue :)

PR #10, opened 2020-01-05
Fix typo

validatedev:


AdnanHodzic:
Hi @validatedev thank you for contribution, but actually tool is called "auto-cpufreq". Hence, in this case changing to from  "auto-cpufreq" to "autocpu-freq" would be making a typo :)

Sorry, but this is not going to be merged.

validatedev:
Well, the command on the readme doesn’t work. Could you have a chance for looking at it?

validatedev:
Oh, wrong change :/

PR #11, opened 2020-01-05
Fix typo

validatedev:
Well, this is the real typo I tried to fix.


AdnanHodzic:
Great, thank you for catching that!

PR #19, opened 2020-01-06
Use nproc in cpufreqctl.sh

johnathana:
Use **nproc** tool to get the number of processing units available. It is part of the GNU coreutils and comes preinstalled in all modern linux distributions.


AdnanHodzic:
LGTM, thank you for your contribution.

PR #24, opened 2020-01-08
temp fix for https://github.com/giampaolo/psutil/issues/1658

validatedev:
There is a bug on psutil 5.6.7, reported at https://github.com/giampaolo/psutil/issues/1658 . Until that issue is resolved, this patch should help the fix the problem temporarily.


validatedev:
The problem was on my battery, investigating.

PR #28, opened 2020-01-09
set_powersave performance enhancement

e4BdSBmUzHowFico5Ktn:
now set_powersave only turns on turbo boost when all 1m load avg is higher than all available cores instead of 4 cores, this should work better for a broader range of cpus.


AdnanHodzic:
Makes sense. Might be a smart thing to do same or similar thing for performance mode as well.

````
    # conditions for setting turbo in performance
    if load1m > 1:
````

Regardless, thank you for your contribution!

PR #34, opened 2020-01-12
fix low performance when charging

validatedev:
I think there isn't any necessity for limiting the performance of CPU when the PC on charging status. In case of config file improvement, the if-else block isn't touched.

### DIFF
--- a/auto-cpufreq.py
+++ b/auto-cpufreq.py
@@ -167,7 +167,7 @@ def set_powersave():
         print("High load, setting turbo boost: on")
         s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
         footer(79)
-    elif cpuload > 40 and cpuload <= 65:
+    elif cpuload > 25 and cpuload <= 60:
         print("Setting to use: \"powersave\" governor")
         s.run("cpufreqctl --governor --set=powersave", shell=True)
 
--- a/auto-cpufreq.py
+++ b/auto-cpufreq.py
@@ -159,7 +159,7 @@ def set_powersave():
         print("High load, setting turbo boost: on")
         s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
         footer(79)
-    elif cpuload > 40 and cpuload <= 70:
+    elif cpuload > 40 and cpuload <= 65:
         print("Setting to use: \"powersave\" governor")
         s.run("cpufreqctl --governor --set=powersave", shell=True)
 
@@ -167,7 +167,7 @@ def set_powersave():
         s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
         #print("\n" + "-" * 60 + "\n")
         footer(79)
-    elif cpuload > 70:
+    elif cpuload > 65:
         print("Setting to use: \"performance\" governor")
         s.run("cpufreqctl --governor --set=performance", shell=True)
 
--- a/auto-cpufreq.py
+++ b/auto-cpufreq.py
@@ -437,7 +437,7 @@ def cli(monitor, live, daemon, install, log):
                 sysinfo()
                 set_autofreq()
                 countdown(5)
-                subprocess.call("clear")
+                s.call("clear")
         elif monitor:
             while True:
                 running_check()
@@ -447,7 +447,7 @@ def cli(monitor, live, daemon, install, log):
                 sysinfo()
                 mon_autofreq()
                 countdown(5)
-                subprocess.call("clear")
+                s.call("clear")
         elif live:
             while True:
                 running_check()
@@ -457,7 +457,7 @@ def cli(monitor, live, daemon, install, log):
                 sysinfo()
                 set_autofreq()
                 countdown(5)
-                subprocess.call("clear")
+                s.call("clear")
         elif log:
                 read_log()
         elif install:
--- a/auto-cpufreq.py
+++ b/auto-cpufreq.py
@@ -145,18 +145,11 @@ def set_powersave():
     print("Total system load:", load1m, "\n")
 
     # conditions for setting turbo in powersave
-    if load1m >= cpus:
+    if load1m >= cpus * (2 / 3):
         print("Setting to use: \"performance\" governor")
         s.run("cpufreqctl --governor --set=performance", shell=True)
 
         print("High load, setting turbo boost: on")
-        s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
-        footer(79)
-    elif load1m > cpus / 6 and load1m < cpus:
-        print("Setting to use: \"powersave\" governor")
-        s.run("cpufreqctl --governor --set=powersave", shell=True)
-
-        print("Medium load, setting turbo boost: on")
         s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
         footer(79)
     elif cpuload > 65:
@@ -166,6 +159,13 @@ def set_powersave():
         print("Very high CPU load, setting turbo boost: on")
         s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
         #print("\n" + "-" * 60 + "\n")
+        footer(79)
+    elif load1m > cpus / 6 and load1m < cpus * (2 / 3):
+        print("Setting to use: \"powersave\" governor")
+        s.run("cpufreqctl --governor --set=powersave", shell=True)
+
+        print("Medium load, setting turbo boost: on")
+        s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
         footer(79)
     elif cpuload > 40 and cpuload <= 65:
         print("Setting to use: \"powersave\" governor")
--- a/auto-cpufreq.py
+++ b/auto-cpufreq.py
@@ -145,7 +145,7 @@ def set_powersave():
     print("Total system load:", load1m, "\n")
 
     # conditions for setting turbo in powersave
-    if load1m > cpus and load1m < cpus * 2:
+    if load1m > 1 and load1m < cpus * 2:
         print("Setting to use: \"powersave\" governor")
         s.run("cpufreqctl --governor --set=powersave", shell=True)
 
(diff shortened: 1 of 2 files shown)


AdnanHodzic:
I introduced this change recently, as at the beginning I was of same opinion as you.

This was [initially suggested](https://www.reddit.com/r/linux/comments/ejxx9f/github_autocpufreq_automatic_cpu_speed_power/fd5nodm/) on Reddit, and thinking about it made a lot sense. Hence I made this changes as first step towards this goal. 

Because if you have a really low load, why do you need CPU running at its absolute max. All it'll do in this case is save power/electricity used which isn't much on a single computer. But image how much power could be saved if `auto-cpufreq` was running in a datacenter on thousands of servers, in that case that energy save could be very valuable.

validatedev:
I'll test some changes and inform you. With that type of change that has been introduced by you, I haven't a good relationship with GNOME, it stutters and lags.

validatedev:
Continuing the test of that PR, and will make some changes when necessary. If you have a better idea, share with me. There is a really performance drop when using that DE with your changes.

AdnanHodzic:
> Continuing the test of that PR, and will make some changes when necessary. If you have a better idea, share with me. There is a really performance drop when using that DE with your changes.

This is good, exactly what I was also thinking at this point. I'll also do some testing using these change and will give you my feedback.

I also use GNOME, and I have 0 performance issues with existing implementation in GNOME or anywhere else.

auto-cpufreq.py
```diff
@@ -221,8 +221,8 @@ def set_performance():
         #print("\n" + "-" * 60 + "\n")
         footer(79)
     else:
-        print("Load optimal, setting turbo boost: off")
-        s.run("echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
+        print("Load optimal, setting turbo boost: on")
+        s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
         #print("\n" + "-" * 60 + "\n")
```
AdnanHodzic: If you put "turbo on" in else statement, then turbo will never turn off in performance, which is contradicting with [what I said in my comment to this MR](https://github.com/AdnanHodzic/auto-cpufreq/pull/34#issuecomment-573408002).

auto-cpufreq.py
```diff
@@ -210,7 +210,7 @@ def set_performance():
     print("\nTotal CPU usage:", cpuload, "%")
     print("Total system load:", load1m, "\n")
 
-    # conditions for setting turbo in performance
+    # conditions for setting turbo in performance for future using
     if load1m > 1:
```
AdnanHodzic: What did you add "for future using"?

validatedev: This comes with the first commit of that PR, undoing the change.

auto-cpufreq.py
validatedev: But I changed the governor "performance" to "powersave". powersave governor will not push the clock to turbo state if it isn't really a necessity but I really don't sure which one gives the good performance. My laptop has 1.8 GHz without TB, 3.9 with TB.

validatedev:
Don't sure the PR https://github.com/AdnanHodzic/auto-cpufreq/pull/28 is a good solution. What `os.getloadavg()` function actually take?

validatedev:
`load1m` cannot exceed any time `cpus`, so I did some adjustments for accurate switching mechanism.

AdnanHodzic:
Sorry for delayed response from my side, been very busy last couple of days. I'll thoroughly review this tomorrow.

validatedev:
Take your time, I’m also experimenting the variables. Also thinking to implement the governor-turbo boost changing based on temperature, but, well I think this should not be belong there, I mean this PR.

auto-cpufreq.py
```diff
@@ -148,24 +145,47 @@ def set_powersave():
     print("Total system load:", load1m, "\n")
 
     # conditions for setting turbo in powersave
-    if load1m > cpus:
+    if load1m >= cpus * (3 / 5):
+        print("Setting to use: \"performance\" governor")
+        s.run("cpufreqctl --governor --set=performance", shell=True)
+
```
AdnanHodzic: I'm not sure about switching to "performance" while in "powersave" mode, regardless of the load/CPU usage. 

Only thing that would give us valuable metric would be to test how much battery would be used on battery while in "peformance" mode on battery vs  given same conditions in "powersave". I believe more battery will be used in "performance".

Hence the reason why I went with option to turn turbo boost on instead, which should handle the load.

Also, based on what criteria did you come up with this number: `load1m >= cpus * (3 / 5)`

auto-cpufreq.py
AdnanHodzic: I made another comment regarding setting switching to `performance` while in `powersave` (discharging battery).

Please correct me if I'm wrong, but you're saying while in "powersave", turbo never goes on for you? Because for me it certainly does.

auto-cpufreq.py
validatedev: I just want to sure about the user's CPU load is more than 'more than the average'. Because if there is a very high load, even though the discharging, we should provide to the user the `performance` governor. The `3/5` is changeable though.

auto-cpufreq.py
validatedev: I'm not saying that. When on the charging mode we will run `set_performance()` function. If there isn't any or little load of CPU, we should provide to the user `powersave` governor because we don't need the high clock speed, but if there is a need for turboboost, `powersave` will definitely handle that, and helps to the PC for staying cool. Beside of that, if there is a noticeable load, we give to the user the power and activate `performance` governor.

auto-cpufreq.py
AdnanHodzic: After thinking about this for couple of days now, let's do this. Currently `auto-cpufreq` is really "trigger ready". In sense that it will easily switch governors and turn on/off turbo.

I like what you did here with calculations, but instead of switching to "performance" governor while on battery, let it turn on/off turbo instead depending on load/cpu usage while in powersave (battery discharging). 

This way we'll have refined way of turning turbo on/off and tool being less of "trigger friendly".

validatedev: @AdnanHodzic I made some changes about your explanation. Could you have a chance for looking at it?

AdnanHodzic: Right, but currently as part of `set_performance` function it's still set to change governors, which is not good.

```
        print("Setting to use: \"powersave\" governor")
        s.run("cpufreqctl --governor --set=powersave", shell=True)

        print("Load optimal, setting turbo boost: on")
        s.run("echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo", shell=True)
```

Which results in situations like these:

```
Battery is: charging

Total CPU usage: 2.0 %
Total system load: 0.24 

Load optimal, setting turbo boost: on
```

Why is turbo on if load is optimal? Ass I said, let's not change governors if it's charging (performance) or discharging (powersave).

Once in one of those 2 governors, let's just turn turbo on/off based on more refined cpu usage/system load.

auto-cpufreq.py
AdnanHodzic: As I discussed in [my other comment](https://github.com/AdnanHodzic/auto-cpufreq/pull/34#discussion_r367795655), at this stage we really shouldn't change "governors" based on cpu usage/system load.

I would rather refine that turbo is turned on/off based on these in a more refined way. Let's take care of the other comment I made and merge this PR.

validatedev:
About the comment that you did, sorry for the late reply. I have been very busy recently and as soon as I am free, I’ll be back there.

AdnanHodzic:
> About the comment that you did, sorry for the late reply. I have been very busy recently and as soon as I am free, I’ll be back there.

No worries, take your time.

auto-cpufreq.py
validatedev: Well, I think if there is no or small load of CPU, pushing the clock to the best base clock freq is not necessary, which is performance governor does. If we are very keen on power saving when low load, I think that way is more accurate because powersave governor could lower the freq of the CPU and if necessary, it will push the clock freq to the higher.

AdnanHodzic: I'm sorry, but I don't agree. There is "small load", let's say you're on battery power and hence in powersave, you're watching 1 minute 4K Youtube video. Turbo kicks in, does the heavy load once the load is gone, it turns off turbo. Battery was preserved to its max, while CPU temperature/fans stayed at their optimal levels.

Only exception, where I agree with you that app should switch from powersave to performance (while on battery) is if we could figure out some kind of algorithm or equation where it detects it's reallynecessary. For example, if load and cpu usage (and in future CPU temp/fan speed) are running high for 5 minutes, besides turbo also switch the governor form powersave to performance.

Otherwise if you have app switching to performance while on battery battery for every little thng, battery life will suffer just as it would without this tool, which then defeats its purpose.

But, these are things that can be done at later stages. Right now, I think it would be the best if we could refine the current settings under which "turbo" is turned on/off and make the app less "trigger friendly", which you already did with i.e: ` if load1m > cpus / 7:`

But let's do it once step at a time is what I'm suggesting here.

validatedev: > For example, if load and cpu usage (and in future CPU temp/fan speed) are running high for >= 5 minutes, besides turbo also switch the governor from powersave to performance.

That looks easily doable.

Other than, I have already reverted what did you say of "performance governor when battery" situation. I think you understand my last comment wrong. I said these words for "powersave governor with turbo boost capability -it doesn't mean we push the clock to highest- with when charging". Could you have a chance for giving an explanation for your thoughts? Why performance with turbo boost capability off (that means we are pushing the clock to base speed which is the 1.6 GHz for my case) instead of powersave with turboboost capability enabled (no low load -> 0.8-1.2 GHz, if load it increases but it drops quickly when not necessary)?

AdnanHodzic:
@validatedev maybe changes we agree on from this PR should also go as part of [snap branch](https://github.com/AdnanHodzic/auto-cpufreq/tree/snap), just as it was the case with #40 

I would really like to see the ["calculations" part](https://github.com/AdnanHodzic/auto-cpufreq/pull/34/files#diff-1dc7576a864b4767103cb92b6df3f9c9R148) to also be merged, as they make the tool less "trigger friendly". 

Basically it's all good, except changing the governor to `performance` while in `powersave` mode. And as I mentioned, even these changes can be made but let's do it at later stage.

Think about it and if you have time please feel free to make a new MR, there's still time as I'm waiting for some interface related changes to be made from Ubuntu side.

validatedev:
@AdnanHodzic OK, I'll make these changes that you're referring today. Closing this PR.

Issue #31, opened 2020-01-11
auto-cpufreq --log seems stopped after resume


validatedev:
Cannot reopen https://github.com/AdnanHodzic/auto-cpufreq/issues/6 myself, so opening the issue again. After suspending and resuming the laptop causes stopping `auto-cpufreq --log` output. It freezes.

validatedev:
Output of `systemctl status auto-cpufreq`
```
● auto-cpufreq.service - auto-cpufreq - Automatic CPU speed & power optimizer f>
     Loaded: loaded (/etc/systemd/system/auto-cpufreq.service; enabled; vendor >
     Active: active (running) since Sat 2020-01-11 11:34:22 +03; 52min ago
   Main PID: 1811 (python3)
      Tasks: 1 (limit: 9195)
     Memory: 17.8M
     CGroup: /system.slice/auto-cpufreq.service
             └─1811 /usr/bin/python3 /usr/bin/auto-cpufreq --daemon

```

validatedev:
FYI, `sudo python3 auto-cpufreq.py --log` also seems stopped.

Exorust:
This is possibly because of saving the .service file in /lib directory instead of the /usr/lib directory.

validatedev:
> This is possibly because of saving the .service file in /lib directory instead of the /usr/lib directory.

Do you have have a source for that, or just an intuition?

Exorust:
Unfortunately only an intuition, I can't track down where the log is being saved. The logging info is being saved. I can verify that through, `systemctl status auto-cpufreq -n20`. But still `auto-cpufreq --log` doesn't work. 

Apart from that, I have also experimented with changing the `service` file and making the following edits,
```
StandardOutput=append:/var/log/auto-cpufreq.log
```
Still no result.

AdnanHodzic:
@Exorust can you update me [with what I suggested](https://github.com/AdnanHodzic/auto-cpufreq/issues/6#issuecomment-573389224) in #6

Exorust:
Its empty

AdnanHodzic:
Then something is terribly wrong. Can you tell me what distro are you using, along with Python version and version for psutil, click, power packages.

validatedev:
That command sometimes freezes, sometimes works. Interesting.

validatedev:
We can do something with the changes on the .py file. Will trying to redirect the stdout to log file directly.

validatedev:
[...]

PR #35, opened 2020-01-12
fix auto-cpufreq --log freezing

validatedev:
This PR is supposed to fix https://github.com/AdnanHodzic/auto-cpufreq/issues/31 .

### DIFF
--- a/scripts/auto-cpufreq.service
+++ b/scripts/auto-cpufreq.service
@@ -6,7 +6,7 @@ ConditionPathExists=/var/log/auto-cpufreq.log
 [Service]
 Type=simple
 User=root
-ExecStart=/usr/bin/sh -c '/usr/bin/python3 /usr/bin/auto-cpufreq --daemon'
+ExecStart=/usr/bin/python3 /usr/bin/auto-cpufreq --daemon
 StandardOutput=append:/var/log/auto-cpufreq.log
 [Install]
 WantedBy=multi-user.target
--- a/scripts/auto-cpufreq.service
+++ b/scripts/auto-cpufreq.service
@@ -6,7 +6,7 @@ ConditionPathExists=/var/log/auto-cpufreq.log
 [Service]
 Type=simple
 User=root
-ExecStart=/usr/bin/python3 /usr/bin/auto-cpufreq --daemon
-StandardOutput=file:/var/log/auto-cpufreq.log
+ExecStart=/usr/bin/sh -c '/usr/bin/python3 /usr/bin/auto-cpufreq --daemon'
+StandardOutput=append:/var/log/auto-cpufreq.log
 [Install]
 WantedBy=multi-user.target


validatedev:
Testing for hours and didn't encounter any issue. All goes well.

AdnanHodzic:
Sorry for delayed response from my side, been very busy last couple of days. I'll thoroughly review this tomorrow.

AdnanHodzic:
Just to confirm, did you also have the problem with logs freezing?

Because no matter what I do without and with these changes my logs never freeze, and I can see them after suspend/hibernate. So I'm not sure how these changes fix this issue since I can't reproduce reported problem in #31 

Regarding changes you made:
```
ExecStart=/usr/bin/sh -c '/usr/bin/python3 /usr/bin/auto-cpufreq --daemon'
StandardOutput=append:/var/log/auto-cpufreq.log
```

During development of this tool, at one point I had these same changes. Reason why I went the path I went with:

```
ExecStart=/usr/bin/python3 /usr/bin/auto-cpufreq --daemon
StandardOutput=file:/var/log/auto-cpufreq.log
```

1. Process is run directly by systemd, no need to create a bash wrapper around it. Potentially having smaller memory footprint

2. These no need for these logs to be "appended", as the file can be overwritten each time. Also potentially resulting in generation of big log size and then needing log retention. Altough at first look this doesn't seem to be the case.

So if you could share your thinking for number 1 and 2 that would be great.

In meantime, let's see if anyone from #31 reports if this fixed their problem.

validatedev:
@AdnanHodzic 
    1. When I try the changes without the bash wrapper, it works as before. Because of that, I share the same thoughts with you and I'll remove the changes for `ExecStart`.
    2. You're right for the log size, we cannot eliminate of that. Maybe we should introduce the internal changes instead of changing the service file. Moreover, if the user decide to delete the log, instead of the display warning to the user, we should recreate the log file and continue.

AdnanHodzic:
Let's leave this PR parked for now. I want people from #31 to confirm this fixes their issue. 

As I cannot replicate the problem they are reporting and in case we pull this PR with master, we only might introduce new problems without being 100% sure it addresses the one it was made for.

validatedev:
I tested for days and didn't encounter any weird behaviour, just for your info.

AdnanHodzic:
During my testing, I also didn't encounter any issues. So you did a good job, I'm just saying it still leaves us with:

> 2. These no need for these logs to be "appended", as the file can be overwritten each time. Also potentially resulting in generation of big log size and then needing log retention. Although at first look this doesn't seem to be the case.

And I would truly like to understand why #31 happens for some people, so until get more data/input let's leave it as is.

AdnanHodzic:
@validatedev I want to merge these changes, since last couple of weeks I've been working what will ultimately become v1.0 of auto-cpufreq. 

Could you please add these changes as [part of snap branch](https://github.com/AdnanHodzic/auto-cpufreq/tree/snap) as that branch has many changes and will be merged with master in next couple of days.

Thanks!

validatedev:
Then I'll close that PR and open the new.

PR #40, opened 2020-02-16
Fix file log freezing issue

validatedev:
That PR is supposed to fix https://github.com/AdnanHodzic/auto-cpufreq/issues/31 .


AdnanHodzic:
Great, thanks

PR #41, opened 2020-02-16
Change some parameters for better use

validatedev:
This PR is successor of the https://github.com/AdnanHodzic/auto-cpufreq/pull/34 .


validatedev:
Side note: I haven't tried these changes (don't using GNU/Linux atm), it should work because changes are the same with the ancestor, but you'd better try before merging.

PR #44, opened 2020-03-19
Add instruction for preventing the conflict with TLP

validatedev:
If you don't like the way of writing in terms of the integrity of the README, please tell me.


AdnanHodzic:
IMO the best way to handle this would be to make code changes to detect if TLP is installed and running. After `auto-cpufreq` install, instruct user they should add these changes. If you want please open a new issue and name it "[feature request] TLP detection and instruction" and I'll look at with I have time.

Otherwise, you can give it a try yourself :)

Other way would be to extend README and I'm not big fan of doing this.

validatedev:
@AdnanHodzic 
Seems reasonable.
I'll try to implement, doesn't look challenging.
Closing the PR.

AdnanHodzic:
@validatedev sounds good! Looking forward to your PR and once you have it please give me rights to also make changes if necessary.

Issue #14, opened 2020-01-05
KeyError: 'coretemp'


racuna:
From another directory I got this (file does not exist):
`cp: no se puede efectuar `stat' sobre 'scripts/cpufreqctl.sh': No existe el fichero o el directorio`

But, executing when pwd is inside the directory auto-cpufreq, then the error mensaje is this:

```
python3 auto-cpufreq.py --monitor

----------------------------- System information ------------------------------

Linux distro: PCLinuxOS 2020 PCLinuxOS
Linux kernel: 5.4.8-pclos1
Driver: intel_pstate
Architecture: x86_64
Processor: Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz
Cores: 4

------------------------------ Current CPU states ------------------------------

CPU max frequency: 
3200 MHz

CPU frequency for each core:

CPU0: 1898 MHz
CPU1: 1702 MHz
CPU2: 2149 MHz
CPU3: 1566 MHz

Temperature for each physical core:

Traceback (most recent call last):
  File "auto-cpufreq.py", line 437, in <module>
    cli()
  File "/usr/local/lib64/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib64/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib64/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib64/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "auto-cpufreq.py", line 410, in cli
    sysinfo()
  File "auto-cpufreq.py", line 338, in sysinfo
    print("CPU" + str(core_num) + " temp: {:.0f}".format(core_temp['coretemp'][core_num].current) + "°C")
KeyError: 'coretemp'
```

Lusitaniae:
did you install all dependencies?

`sudo apt install python3 python3-distro python3-psutil python3-click -y`

What system are you using?

magneticleaves:
I'm also getting this error! I've installed all the listed dependencies.

```
$ sudo auto-cpufreq --monitor
----------------------------- System information ------------------------------

Linux distro: Debian GNU/Linux 10 buster
Linux kernel: 4.19.0-6-amd64
Driver: acpi-cpufreq
Architecture: 
[...]

Issue #17, opened 2020-01-06
Error: Couldn't find any of the necessary scaling governors. AMD CPU possibly not supported?


sandersantema:
I got this error while running the tool with a AMD Ryzen 7 PRO 3700U cpu. 

If this error is indeed caused because the AMD cpu isn't supported, might I suggest updating the README to reflect this? If this isn't what's causing the error I'd have to look more into it and I'd gladly provide any needed information.

Output of `sudo tlp-stat -p`
```
--- TLP 1.2.2 --------------------------------------------

+++ Processor
CPU model      = AMD Ryzen 7 PRO 3700U w/ Radeon Vega Mobile Gfx

/sys/devices/system/cpu/cpu0/cpufreq/scaling_driver    = acpi-cpufreq
/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor  = ondemand
/sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors = ondemand performance schedutil 
/sys/devices/system/cpu/cpu0/cpufreq/scaling_min_freq  =  1400000 [kHz]
/sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq  =  2300000 [kHz]
/sys/devices/system/cpu/cpu0/cpufreq/scaling_available_frequencies = 2300000 1700000 1400000 [kHz]

(Idem for the remaining seven cores)

/sys/devices/system/cpu/cpufreq/boost                  = 1

x86_energy_perf_policy: unsupported CPU.

/sys/module/workqueue/parameters/power_efficient       = N
/proc/sys/kernel/nmi_watchdog                          = 0

+++ Undervolting
PHC kernel not available.
```

AdnanHodzic:
When I was making this tool I was strictly targeting Intel, as I didn't have any AMD CPU's to test it. Either way thank you for updating me about this.

As a goal to extend this tool for AMD as well, could you please confirm that in AMD's case, "ondemand" acts like "powersave" for Intel?

And please correct me if I'm wrong, but it seems you have following governors: `ondemand performance schedutil`?

`sudo tlp-stat -p` shows it, but you can also verify by running: `sudo cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors`

sandersantema:
> As a goal to extend this tool for AMD as well, could you please confirm that in AMD's case, "ondemand" acts like "powersave" for Intel?
 
[...]

PR #62, opened 2020-08-03
Refactoring and fix for #14 (AMD issue)

Red-Eyed:
* A bit of refactoring
* solves #14 As I didn't find how to get temperature for AMD Ryzen, I just display `acpitz` instead of core temp.

### DIFF
--- a/bin/auto-cpufreq
+++ b/bin/auto-cpufreq
@@ -6,8 +6,12 @@
 
 # core import
 import sys
+from subprocess import call
+
 sys.path.append('../')
-from source.core import * 
+from source.core import *
+import click
+
 
 # ToD: replace every s.call with s.run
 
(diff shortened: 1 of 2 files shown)


AdnanHodzic:
This is great, thank you for your contribution!

Since it seems you have an AMD CPU and know your way around it. If you have time to look at #17 that would be great. As resolving this would basically add AMD support for `auto-cpufreq` which would be a huge contribution.

Red-Eyed:
Yeah, I read about #17. But in my case , I have 2 governors: `performance` and `powersafe`.  
It seems that some AMD CPUs don't have `powersafe` but `schedutil` instead.

BTW: I increased my bat life from 5h to 9h! it's very nice and helpful project, thank you for this!  
Also, it's cool that it's written in python, because I'm a python dev and very love thins language, so glad to contribute :)

AdnanHodzic:
> Yeah, I read about #17. But in my case , I have 2 governors: performance and powersafe.
> It seems that some AMD CPUs don't have powersafe but schedutil instead.

Weird, first thing that comes to mind is to make additional check which governors to use based on which AMD CPU is detected.

> BTW: I increased my bat life from 5h to 9h! it's very nice and helpful project, thank you for this!

Most welcome, thank you as it's always great to hear feedback like this :)

>  Also, it's cool that it's written in python, because I'm a python dev and very love thins language, so glad to contribute :)

Great! In that case I hope I'll see more of your pull requests and I'll make sure to add your contributions in release notes of a new release.

PR #66, opened 2020-08-05
added issue template

Red-Eyed:


AdnanHodzic:
Great, stuff.

Just so you are aware where I was going with this. For issues, I was planning to implement similar/same feature to "debug" option I created for [displaylink-debian tool.](https://github.com/AdnanHodzic/displaylink-debian).

This way whoever was about to report an issue could run `auto-cpufreq --debug` and include useful info to the bug report, i.e: sys info, governors, etc ...

Red-Eyed:
We can add `inxi` as deps and call `inxi -Fz` on `--debug`
How do u think?

AdnanHodzic:
I think it's a great for now. It (`inxi`) has everything we need, except maybe the available governors.

Red-Eyed:
Yeah governors would be nice to show.

AdnanHodzic:
and useful, as I think a lot of source issues could relate to this. Maybe a mix of `inxi` and `auto-cpufreq` output only showing governor info?

Red-Eyed:
So, we can add `inxi` as a deps, and then implement `--debug` option to format output with governors

AdnanHodzic:
Sounds good to me.

PR #68, opened 2020-08-05
switching to psutil in order to monitor charger state

Red-Eyed:
Suggesting to switch from `power` to `psutil` package in order to monitor battery state.
Reason: `psutil` is very active project compared to  `power`. 

After merging other PR's I'll update it


Red-Eyed:
Merged master, can be reviewed

Issue #59, opened 2020-05-21
Permission denied on Snap 2.44.3+20.04


Booteille:
Hi!

Using `auto-cpufreq --log` after a successful `sudo auto-cpufreq --install`, I get the following error:

```console
⋊> ~ auto-cpufreq --log

-------------------------------- Daemon check ---------------------------------

ERROR:

Daemon not enabled, must run install first, i.e:
sudo auto-cpufreq --install

-------------------------------------------------------------------------------


----------------------------- System information ------------------------------

Traceback (most recent call last):
  File "/snap/auto-cpufreq/5/bin/auto-cpufreq", line 108, in <module>
    main()
  File "/snap/auto-cpufreq/5/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/snap/auto-cpufreq/5/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/snap/auto-cpufreq/5/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/snap/auto-cpufreq/5/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/snap/auto-cpufreq/5/bin/auto-cpufreq", line 39, in main
    sysinfo()
  File "/snap/auto-cpufreq/5/lib/python3.6/site-packages/source/core.py", line 351, in sysinfo
    searchfile = open("/var/lib/snapd/hostfs/etc/os-release", "r")
PermissionError: [Errno 13] Permission denied: '/var/lib/snapd/hostfs/etc/os-release'
```

Tried the command with `sudo`, same result.

AdnanHodzic:
Could you please tell me which distribution/release are you running?

Booteille:
Sure!

Pop_OS! 20.04

AdnanHodzic:
Will look into this. Until I resolve this issue, I would suggest to use [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq#auto-cpufreq-installer)

Booteille:
That's what I finally did.

Thank you!

AdnanHodzic:
I would be interested in hearing if you still have this problem with snap as as part of new (unreleased) release/beta:

```
sudo snap install auto-cpufreq --beta
```

[...]

PR #75, opened 2020-08-06
added hot fix for snap for pop os

Red-Eyed:


AdnanHodzic:
Let's continue discussion from #59 here.

All you need to do to build a snap with current structure is: 

```
snapcraft
```

After build is successful, all you need to do is install the snap, i.e:

```
sudo snap install auto-cpufreq_1.2_amd64.snap --dangerous
```
and then enable necessary interfaces, i.e:

```
sudo snap connect auto-cpufreq:cpu-control
sudo snap connect auto-cpufreq:hardware-observe
sudo snap connect auto-cpufreq:system-observe
```

A lot of people on on forum.snapcraft.com are Canonical employees and are more then willing to help as you [can see from an example](https://forum.snapcraft.io/t/snap-not-buildable-anymore/19274).

For everything else [I suggest you refer to their official documentation](https://snapcraft.io/docs/creating-a-snap) or a ask me a question here :)

Red-Eyed:
Just wonder,  there any specific reasons to add gcc and python devel ?

AdnanHodzic:
> Just wonder, are there any specific reasons to add gcc and python devel ?

They were missing as build dependencies. I could've also used deb packages, i.e `python3-psutils` but it's much older version then when downloaded using `pip3` which is what's used in this case.

What you did in this PR is good, but worst case it won't display distro info. Regardless, I pushed it to `edge` channel so feel free to test it by running:

```
sudo snap install auto-cpufreq --edge
```

Red-Eyed:
Yes, it works.
But before after merging, merge please this #76

Issue #84, opened 2020-08-07
FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/auto-cpufreq/scripts/auto-cpufreq-install.sh'


Bu4ak:
```bash
yay -S auto-cpufreq
sudo auto-cpufreq --install
```
### Error output:
```text
--------------------- Deploying auto-cpufreq as a daemon ----------------------

* Turn off bluetooth on boot

* Deploy auto-cpufreq install script
Traceback (most recent call last):
  File "/usr/bin/auto-cpufreq", line 115, in <module>
    main()
  File "/usr/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/bin/auto-cpufreq", line 98, in main
    deploy_daemon()
  File "/usr/lib/python3.8/site-packages/source/core.py", line 215, in deploy_daemon
    shutil.copy(SCRIPTS_DIR / "auto-cpufreq-install.sh", "/usr/bin/auto-cpufreq-install")
  File "/usr/lib/python3.8/shutil.py", line 415, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.8/shutil.py", line 261, in copyfile
    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/auto-cpufreq/scripts/auto-cpufreq-install.sh'

```
---

### System information:
```text
-------------------------------------------------------------------------------

System: Hewlett-Packard product: HP EliteBook Folio 1040 G1 v: A3009D510203 
           serial: <filter> 
           Mobo: Hewlett-Packard model: 213E v: KBC Version 24.2B serial: <filter> 
           BIOS: Hewlett-Packard v: L83 Ver. 01.21 date: 08/19/2014 
Battery:   ID-1: BAT0 charge: 25.1 Wh condition: 31.1/31.1 Wh (100%) 
[...]

PR #90, opened 2020-08-10
Reorganize README.md

yochananmarqos:
Install, *then* use. Not the other way around. Now it's clearer for folks using the AUR package they can just start using it. The AUR package install file notifies the user there is a service. Keep It Simple.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,24 @@ In case you encounter any problems with `auto-cpufreq-installer`, please [submit
 
 [AUR package is available](https://aur.archlinux.org/packages/auto-cpufreq-git/) for install. After which `auto-cpufreq` will be available as a binary and you can skip to [auto-cpufreq: modes and options](https://github.com/AdnanHodzic/auto-cpufreq#auto-cpufreq-modes-and-options) for further reference.
 
+### Install - auto-cpufreq daemon
+
+Necessary changes are made to the system for auto-cpufreq CPU optimizaton to persist across reboots. Daemon is deployed and then started as a systemd service. Changes are made automatically and live log is made for monitoring purposes.
+
+`sudo auto-cpufreq --install`
+
+After daemon is installed, `auto-cpufreq` is available as a binary and is running in the background. Its logs can be viewed by running: `auto-cpufreq --log`
+
+Since daemon is running as a systemd service, its status can be seen by running:
+
+`systemctl status auto-cpufreq`
+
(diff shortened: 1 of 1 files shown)


Red-Eyed:
What about help message?
```
auto-cpufreq --help
Usage: auto-cpufreq [OPTIONS]

Options:
  --monitor             Monitor and suggest CPU optimizations
  --live                Monitor and make suggested CPU optimizations
  --install / --remove  Install/remove daemon for automatic CPU optimizations
  --log                 View live CPU optimization log made by daemon
  --debug               Show debug info (include when submitting bugs)
  --help                Show this message and exit.
```

yochananmarqos:
The help message is another issue. This at least make the README clearer.

AdnanHodzic:
I reverted changes made with this PR.

@yochananmarqos with this change we catered whole README to those who installed `auto-cpufreq` using AUR. 

What is described here, doesn't apply to people who used `snap` or `auto-cpufreq-installer`. What is *simple* to you, doesn't mean it applies and is simple in every other use case. 

I think it would be a better idea to extend "[Arch Linux](https://github.com/AdnanHodzic/auto-cpufreq#arch-linux)" section, and say after "AUR package installation ____" fill in the blanks. As from what I see #84 and #91 after AUR installation, `auto-cpufreq --install` will be automatically run and just the `systemctl` needs to be enabled?

If this is true, I didn't know that was the case with AUR package. In my opinion and to keep things consistent as they are with `snap` and `auto-cpufreq-installer`. It would be the best if after AUR package installation, `auto-cpufreq` would be available as a binary, and then you have an option to run it `live` or `install` or whatever you want to do with it.

Let's discuss discuss some more. 

@Red-Eyed for the time being, could you please leave merging PR's to `master` to me? Thanks

crian:
@AdnanHodzic there is no point in making such a AUR package, where you have manually install things. Then you can build the whole thing from source. 

The AUR package installs things into its path so the package manager (pacman) keeps track of these files. For example: if you remove the package all files will also be removed without manual intervention.

AdnanHodzic:
@crian right, okay. In that case as I suggested in #84, let's simply extend README to have what needs to be done when AUR package is installed and that's it.

AdnanHodzic:
[README ](https://github.com/AdnanHodzic/auto-cpufreq)updated, I hope this clarifies any confusion. Please submit a new PR if you think otherwise.

Issue #98, opened 2020-08-29
Problem enabling Turbo on Ryzen CPUS


rrodriguez81:
Hi,

I own an AMD Ryzen 3700U CPU, the turbo isn't enabled when one thread reach 100% of CPU usage. I need at least two of them to reach the 25% of CPU Load.

I fixed it modifying the source code a bit. Since I can create a new branch and push the possible change, I'm attaching the modified file, maybe it can be useful to other users,

Thanks for you good work,
[core.zip](https://github.com/AdnanHodzic/auto-cpufreq/files/5145595/core.zip)

AdnanHodzic:
Great! 

Could you please create a [PR from a fork](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork)? This would help me in review and also to give you give credit for your contribution, thanks.

rrodriguez81:
Done! take a look at it. maybe is an easy but good solution.

Best regards.

AdnanHodzic:
Great, thank you. Let me look into it and test couple of things and I'll get back to you.

AdnanHodzic:
Fixed as part of #99

PR #99, opened 2020-08-29
Modified fixed cpu usage percentage 

rrodriguez81:
Turbo should be enabled when one thread reach 100% of usage, in original version there was a fixed value of 25%, but this only works when there're 4 threads, in CPUs with more threads 

Reference: #98

### DIFF
--- a/source/core.py
+++ b/source/core.py
@@ -303,7 +303,7 @@ def set_powersave():
     if load1m > CPUS / 7:
         print("High load, setting turbo boost: on")
         turbo(True)
-    elif cpuload > 25:
+    elif cpuload > 100 / CPUS:
         print("High CPU load, setting turbo boost: on")
         turbo(True)
     else:
@@ -331,7 +331,7 @@ def mon_powersave():
             print("Currently turbo boost is: off")
         footer()
 
-    elif cpuload > 25:
+    elif cpuload > 100 / CPUS:
         print("High CPU load, suggesting to set turbo boost: on")
         if turbo():
             print("Currently turbo boost is: on")
--- a/source/core.py
+++ b/source/core.py
@@ -9,6 +9,7 @@ import shutil
 import sys
 import time
 import warnings
+from math import isclose
 from pathlib import Path
 from pprint import pformat
 from subprocess import getoutput, call, run
@@ -303,7 +304,7 @@ def set_powersave():
     if load1m > CPUS / 7:
         print("High load, setting turbo boost: on")
         turbo(True)
-    elif cpuload > 100 / CPUS:
+    elif psutil.cpu_percent(percpu=False) >= 25.0 or isclose(max(psutil.cpu_percent(percpu=True)), 100):
         print("High CPU load, setting turbo boost: on")
         turbo(True)
     else:
@@ -331,7 +332,7 @@ def mon_powersave():
             print("Currently turbo boost is: off")
         footer()
 
-    elif cpuload > 100 / CPUS:
+    elif psutil.cpu_percent(percpu=False) >= 25.0 or isclose(max(psutil.cpu_percent(percpu=True)), 100):
         print("High CPU load, suggesting to set turbo boost: on")
         if turbo():
             print("Currently turbo boost is: on")
@@ -364,7 +365,7 @@ def set_performance():
     if load1m >= CPUS / 5:
         print("High load, setting turbo boost: on")
         turbo(True)
-    elif cpuload > 100 / (CPUS * 2):
+    elif psutil.cpu_percent(percpu=False) >= 15.0 or isclose(max(psutil.cpu_percent(percpu=True)), 100):
         print("High CPU load, setting turbo boost: on")
         turbo(True)
     else:
--- a/source/core.py
+++ b/source/core.py
@@ -364,7 +364,7 @@ def set_performance():
     if load1m >= CPUS / 5:
         print("High load, setting turbo boost: on")
         turbo(True)
-    elif cpuload > 20:
+    elif cpuload > 100 / (CPUS * 2):
         print("High CPU load, setting turbo boost: on")
         turbo(True)
     else:


rrodriguez81:
Wait a bit please, I've detected that on charging state could be improved aswell. I'll do another pr in a few minutes.

rrodriguez81:
The commit has joined to previous PR.

AdnanHodzic:
No worries, take your time, I'll thoroughly look into this in the morning.

Red-Eyed:
@AdnanHodzic I checked this PR:
I just loaded my single core on 100% by command: `python -c "while True: pass"` and it switches to turbo

rrodriguez81:
@Red-Eyed how many threads can handle your CPU? is your laptop in charging state? If there are only 4 is normal, but if the CPU can handle more than 4 CPUs is normal that the CPU load doesn't reach 25% (at least on an AMD Ryzen 3700U, because I haven't tested with an Intel CPU).

In new portable AMD CPUs with more than 4C/8T, with an only 25% usage the problem will more noticeable.

Red-Eyed:
Hello @rrodriguez81 
> how many threads can handle your CPU? 
4

> is your laptop in charging state?
Discharging

rrodriguez81:
Hello @Red-Eyed :) ,

Not sure if because the CPU model, but in my case it didn't launch turbo. I used the stress tool for generate a 100% load:

 sudo stress --cpu 1 --timeout 120

Red-Eyed:
Just tried `sudo stress --cpu 1 --timeout 20` and it works like a charm: 20 secs with `turbo boost: on` than it switched to `turbo boost: off`
I have CPU: ` AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx`

Red-Eyed:
BTW, are you able to switch to turbo anyway?

rrodriguez81:
@Red-Eyed with my modified version works fine, with original version the turbo doesn't get activated unless I use two threads, for example:
 sudo stress --cpu 2 --timeout 20

Red-Eyed:
@AdnanHodzic it seems that this PR should use this `psutil.cpu_percent(percpu=True)`
`psutil.cpu_percent(percpu=True)` will return list with cpu load per core

Red-Eyed:
> @Red-Eyed with my modified version works fine, with original version the turbo doesn't get activated unless I use two threads, for example:
> sudo stress --cpu 2 --timeout 20

Yeah, it seems that I got what u are trying to say...

Red-Eyed:
@AdnanHodzic 
I think that this code example should work correct:
```python
if psutil.cpu_percent(percpu=False) >= 25.0 or max(psutil.cpu_percent(percpu=True) >= 99.0):
    turbo(True)
```

AdnanHodzic:
I can also confirm it works as expected on Intel. I.e `sudo stress --cpu 1 --timeout 15` works like a charm on i7-8565U.

> @AdnanHodzic
> I think that this code example should work correct:
> 
> ```python
> if psutil.cpu_percent(percpu=False) >= 25.0 or isclose(max(psutil.cpu_percent(percpu=True), 100)):
>     turbo(True)
> ```

Are you saying to replace (current) i.e:

```
cpuload > 100 / CPUS
```
with:

```
psutil.cpu_percent(percpu=False) >= 25.0
```

and same result will be achieved?

@rrodriguez81 my only question for you is 

```
elif cpuload > 100 / (CPUS * 2):
```

Could you tell me what was the reasoning behind using `CPU * 2`? Simple way to trigger it easier if it's in performance?

Red-Eyed:
@AdnanHodzic actually, could you please remind me at what conditions turbo should be enabled?

rrodriguez81:
@AdnanHodzic , this is just for launch turbo faster when the power supply is connected.

@Red-Eyed don't worry, all constructive ideas are welcomed :) .

rrodriguez81:
@Red-Eyed I think that there is a problem with you piece of code, it returns this error:
`Traceback (most recent call last):
  File "/usr/bin/auto-cpufreq", line 115, in <module>
    main()
  File "/usr/lib/python3.8/site-packages/click/core.py", line 829, in __call__
  File "/usr/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
  File "/usr/lib/python3.8/site-packages/click/core.py", line 610, in invoke
  File "/usr/bin/auto-cpufreq", line 51, in main
    set_autofreq()
  File "/usr/lib/python3.8/site-packages/source/core.py", line 411, in set_autofreq
    set_powersave()
  File "/usr/lib/python3.8/site-packages/source/core.py", line 308, in set_powersave
    elif psutil.cpu_percent(percpu=False) >= 25.0 or isclose(max(psutil.cpu_percent(percpu=True), 100)):
TypeError: '>' not supported between instances of 'int' and 'list'
`

Red-Eyed:
Yep, just a typo
```python
if psutil.cpu_percent(percpu=False) >= 25.0 or isclose(max(psutil.cpu_percent(percpu=True)), 100):
    turbo(True)
```

AdnanHodzic:
> @AdnanHodzic actually, could you please remind me at what conditions turbo should be enabled?

Before this PR (and I think those values were good and we should keep them):

power-save/battery:

```
load1m > CPUS / 7
cpuload > 25
```

performance:

```
load1m >= CPUS / 5
cpuload > 20
```

And I quite like what @rrodriguez81 did, as it's very simple yet powerful solution.

> @AdnanHodzic , this is just for launch turbo faster when the power supply is connected.

No worries, I deleted my commented in meantime since I figured what you did :)

rrodriguez81:
@Red-Eyed your solution works aswell, not sure why but it takes a few seconds to launch turbo.

capture
capture2

@AdnanHodzic regarding your idea about performance mode, if you set a static value, could be difficult to trigger turbo with CPUs with more than 8 threads.

For example: For 8 threads CPUS: 100 / 8 = the turbo is triggered with 12.5 CPU load, but if there is a Ryzen 4800H, with 8 cores and 16 threads, 100 / 16 , 100%  CPU usage of one thread is 6,25% total of total CPU load.

Not sure if I'm explaining correctly (English is not my native language)

Red-Eyed:
> @Red-Eyed your solution works aswell, not sure why but it takes a few seconds to launch turbo.
> 
> capture
> capture2
> 
> @AdnanHodzic regarding your idea about performance mode, if you set a static value, could be difficult to trigger turbo with CPUs with more than 8 threads.
> 
> For example: For 8 threads CPUS: 100 / 8 = the turbo is triggered with 12.5 CPU load, but if there is a Ryzen 4800H, with 8 cores and 16 threads, 100 / 16 , 100% CPU usage of one thread is 6,25% total of total CPU load.
> 
> Not sure if I'm explaining correctly (English is not my native language)

That's why I proposed that code snippet that is not hardcoded to num of threads

AdnanHodzic:
> > @Red-Eyed your solution works aswell, not sure why but it takes a few seconds to launch turbo.
> > capture
> > capture2
> > @AdnanHodzic regarding your idea about performance mode, if you set a static value, could be difficult to trigger turbo with CPUs with more than 8 threads.
> > For example: For 8 threads CPUS: 100 / 8 = the turbo is triggered with 12.5 CPU load, but if there is a Ryzen 4800H, with 8 cores and 16 threads, 100 / 16 , 100% CPU usage of one thread is 6,25% total of total CPU load.
> > Not sure if I'm explaining correctly (English is not my native language)
> 
> That's why I proposed that code snippet that is not hardcoded to num of threads

Agreed, having it hard coded is terrible. If it's fine with you @rrodriguez81 let's go with @Red-Eyed eye proposal as it's very clear what's happening. But since this your PR I'd like you to commit those changes, let's just figure out this first:

> @Red-Eyed your solution works aswell, not sure why but it takes a few seconds to launch turbo.

rrodriguez81:
@AdnanHodzic @Red-Eyed not sure why, but with 100 / CPUS the turbo get triggered faster than with @Red-Eyed ' s provided solution.

With my solution the turbo is triggered instantly, and with @Red-Eyed sometimes takes a few seconds (3 or 4 seconds not too much). Maybe because the isclose()  is an approximation or only needs an higher value.

But I'm agree using @Red-Eyed if is more clear for the code.

Red-Eyed:
> @AdnanHodzic @Red-Eyed not sure why, but with 100 / CPUS the turbo get triggered faster than with @Red-Eyed ' s provided solution.
> 
> With my solution the turbo is triggered instantly, and with @Red-Eyed sometimes takes a few seconds (3 or 4 seconds not too much). Maybe because the isclose() is an approximation or only needs an higher value.
> 
> But I'm agree using @Red-Eyed if is more clear for the code.

Maybe `psutil` takes such a delay
regarding `isclose` it's just approximation as you [cannot strongly compare floats](https://floating-point-gui.de/errors/comparison/)
instead of `isclose`, u can try
```python
if psutil.cpu_percent(percpu=False) >= 25.0 or max(psutil.cpu_percent(percpu=True) >= 99:
    turbo(True)
```

Red-Eyed (approved):
it seems fine.

rrodriguez81:
@Red-Eyed the same problem, it trigger turbo slower than the 100 / CPUS solution. It takes to me a few seconds aswell.

Red-Eyed:
> @Red-Eyed the same problem, it trigger turbo slower than the 100 / CPUS solution. It takes to me a few seconds aswell.

imho, i don't think that a few seconds is a problem..

Red-Eyed:
measured the time, and it's okay
image

Red-Eyed:
so it seems that `psutil` takes some average that requires some time

rrodriguez81:
A few seconds modifying a core speed to 2.3Ghz o 3.7Ghz for example, could be the duration of the process that get the 100% of a CPU thread. The more dynamic the turbo is launched, the more agile is the notebook.

AdnanHodzic:
> it seems fine.

Also seems fine on Intel. 

> so it seems that `psutil` takes some average that requires some time

This is also fine as I also want to switch so snap using latest Python and Ubuntu 20.04 as base in near future which might also fix things + one of the future psutil lib updates might also address this.

Either way, thank you both and especially to @rrodriguez81 for your first contribution :partying_face:

Red-Eyed:
> A few seconds modifying a core speed to 2.3Ghz o 3.7Ghz for example, could be the duration of the process that get the 100% of a CPU thread. The more dynamic the turbo is launched, the more agile is the notebook.

Just wonder are you sure that switching turbo with high frequency is energy efficient or healthy idk?

Red-Eyed:
interesting, which version of `psutil` do u use?
`python -c 'import psutil; print(psutil.__version__)'`
mine is `5.7.2`

rrodriguez81:
@Red-Eyed I think that in the actual stable version, it tries to launch turbo as soon as is needed, but it didn't have in mind all kinds or CPUs. But if you think that is a good idea to delay it a bit to be more power efficent is ok for me aswell.

The psutil version I'm using is the same as yours, I'm using latest Linux Manjaro distro.
 
@AdnanHodzic thanks to you and all the project contributors, projects like this makes LInux better :) .

AdnanHodzic:
Before I merge this with `master`, one last observation, for me on Intel (i7-8565U ) with psutil 5.7.2.

```
stress --cpu 8 --io 4 --vm 4 --timeout 15
```

On performance will trigger turbo in a 1 or less, which is good. 

```
sudo stress --cpu 1 --timeout 15
```

On performance it can take up to >= 5 seconds to trigger turbo, this is expected now right?

Red-Eyed:
I found one more parameter
@AdnanHodzic would be nice if you try it.
`psutil.cpu_percent(percpu=False, interval=0.01)` so, just add `interval=0.01` to each `psutil.cpu_percent` call and take a look

AdnanHodzic:
> I found one more parameter
> @AdnanHodzic would be nice if you try it.
> `psutil.cpu_percent(percpu=False, interval=0.01)` so, just add `interval=0.01` to each `psutil.cpu_percent` call and take a look

No this makes it worse, as turbo boost is too "trigger happy", it's almost on all the time in both performance and powersave for me.

Let's leave it as it is for now and then we can improve it later on. Merging ...

Red-Eyed:
Okay

PR #105, opened 2020-09-02
Improved turbo trigger on 1 core 100% load and on load average >1m wh…

rrodriguez81:
I've included last @Red-Eyed code suggestion, and also included an optimization for detect correct load when there are more than 8 CPUS (now based on percentage).

I've checked the results with geekbench with charging state and battery. Turbo gets enabled and disabled fast (some times when the system get very high load takes a bit to get disabled).

Based on geekbench results, I'm getting very good results: https://browser.geekbench.com/v5/cpu/search?page=1&q=3700u&utf8=%E2%9C%93

batterymode
performance

Please take a look, maybe some parameters could be improved a bit, but for me works really well.

### DIFF
--- a/source/core.py
+++ b/source/core.py
@@ -301,10 +301,10 @@ def set_powersave():
     print("Total system load:", load1m, "\n")
 
     # conditions for setting turbo in powersave
-    if load1m > CPUS / 7:
+    if load1m > (15*CPUS)/100: 
         print("High load, setting turbo boost: on")
         turbo(True)
-    elif psutil.cpu_percent(percpu=False) >= 25.0 or isclose(max(psutil.cpu_percent(percpu=True)), 100):
+    elif psutil.cpu_percent(percpu=False, interval=0.01) >= 25.0 or isclose(max(psutil.cpu_percent(percpu=True, interval=0.01)), 100):
         print("High CPU load, setting turbo boost: on")
         turbo(True)
     else:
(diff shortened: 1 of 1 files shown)


source/core.py
```diff
@@ -362,10 +361,10 @@ def set_performance():
     print("\nTotal CPU usage:", cpuload, "%")
     print("Total system load:", load1m, "\n")
 
-    if load1m >= CPUS / 5:
+    if load1m >= (10*CPUS)/100:
```
Red-Eyed, line 364: Could someone explain me, what does `load1m` mean?

Red-Eyed (changes requested):

rrodriguez81:
Is the load average in last minute.

source/core.py
AdnanHodzic: "_It conventionally appears in the form of three numbers which represent the system load during the last one-, five-, and fifteen-minute periods._" - Reference: https://en.wikipedia.org/wiki/Load_(computing)

`load1m` var is ought to represent last minute.

AdnanHodzic:
@rrodriguez81 good stuff, thanks for your contribution!

source/core.py
Red-Eyed: oh, I see its weighted average of system load with window size of 1 minute. Thanks

Issue #115, opened 2020-09-16
Report cpu info formated as a table


Haptein:
Cpu info occupies too much space, specially with high core counts.

AdnanHodzic:
Addressed as part of #116

Issue #114, opened 2020-09-16
Wrong core temperatures on 'coretemp'


Haptein:
CPU0's temp corresponds to package temp, CPU1's temp to CPU2's so on and so forth.

Linux distro: Manjaro Linux 20.1 Mikah
Linux kernel: 5.8.6-1-MANJARO
Processor: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Cores: 12
Architecture: x86_64
Driver: intel_pstate

CPU0 temp: 35°C
CPU1 temp: 33°C
CPU2 temp: 34°C
CPU3 temp: 32°C
CPU4 temp: 33°C
CPU5 temp: 33°C

From psutil:
'coretemp': [
shwtemp(label='Package id 0', current=35.0, high=100.0, critical=100.0), 
shwtemp(label='Core 0', current=33.0, high=100.0, critical=100.0),
shwtemp(label='Core 1', current=34.0, high=100.0, critical=100.0),
shwtemp(label='Core 2', current=32.0, high=100.0, critical=100.0),
shwtemp(label='Core 3', current=33.0, high=100.0, critical=100.0),
shwtemp(label='Core 4', current=33.0, high=100.0, critical=100.0),
shwtemp(label='Core 5', current=33.0, high=100.0, critical=100.0),
shwtemp(label='Package id 0', current=35.0, high=100.0, critical=100.0),
shwtemp(label='Core 0', current=33.0, high=100.0, critical=100.0),
shwtemp(label='Core 1', current=34.0, high=100.0, critical=100.0),
shwtemp(label='Core 2', current=32.0, high=100.0, critical=100.0),
shwtemp(label='Core 3', current=33.0, high=100.0, critical=100.0),
shwtemp(label='Core 4', current=33.0, high=100.0, critical=100.0),
shwtemp(label='Core 5', current=33.0, high=100.0, critical=100.0)]

AdnanHodzic:
Addressed as part of #116

PR #116, opened 2020-09-16
Report core states as a table

Haptein:
Tried addressing #115  and #114 (since getting the table data required modifying the same code).
Please corroborate correct reporting of cpu temps, I can only test with my lap cpu.
Btw, is the try/exec block really needed? Are there cases where psutil.sensor_temperatures() reports a shwtemp object without the _current_ property?

### DIFF
--- a/source/core.py
+++ b/source/core.py
@@ -469,7 +469,10 @@ def sysinfo():
             line = f.readline()
 
     # get cores count
-    cpu_count = psutil.cpu_count()
+    cpu_count = psutil.cpu_count(logical=True)
+    phys_cpu_count = psutil.cpu_count(logical=False)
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
> Tried addressing #115 and #114 (since getting the table data required modifying the same code).

Thanks!

> Please corroborate correct reporting of cpu temps, I can only test with my lap cpu.

Yes, it works and looks great on Intel i7. @Red-Eyed can you confirm same case is with AMD?

> Btw, is the try/exec block really needed? Are there cases where psutil.sensor_temperatures() reports a shwtemp object without the _current_ property?

TBH, I can't remember why this `try/except` block is there, but there must be a reason. If you think there's a better way to go about this, please feel free to address it in another PR :)

PR #122, opened 2020-09-24
add support for Solus

natri23:
add an instruction command to point auto-cpufreq.service to the right location to fix --install on Solus


AdnanHodzic:
Changes look good to me. Could you just remove installation of `inxi` package, as with new revamped `--debug` there's no need for it anymore.

PR #123, opened 2020-09-24
added required build-packages for Solus

natri23:
Tested with a fresh install. Solus required system.devel for building packages


AdnanHodzic:
Good catch

Issue #108, opened 2020-09-06
Battery mode is "charging" even unplugged


tibert01:
Hello, I installed auto-cpufreq from snap and by cloning from git. I tried to use the monitor option, but the battery mode rests on "charging" even after unplugging my laptop.
I'm on Pop!_OS, with the latest updates. The kernel version is 5.4.0-7642-generic.

### System information:
```text
System: HP product: HP Pavilion Laptop 15-ck0xx 
           v: Type1ProductConfigId serial: <filter> 
           Mobo: HP model: 841C v: 64.14 serial: <filter> UEFI: Insyde v: F.12 
           date: 12/17/2019 
Battery:   ID-1: BAT0 charge: 27.5 Wh condition: 30.1/30.1 Wh (100%) 
CPU:       Governors: performance powersave Topology: Quad Core model: Intel Core i7-8550U bits: 64 type: MT MCP 
           L2 cache: 8192 KiB 
           Speed: 700 MHz min/max: 400/4000 MHz Core speeds (MHz): 1: 700 2: 701 3: 700 
           4: 700 5: 700 6: 700 7: 700 8: 700 
Graphics:  Device-1: Intel UHD Graphics 620 driver: i915 v: kernel 
           Device-2: NVIDIA GP108M [GeForce MX150] driver: nvidia v: 440.100 
           Display: x11 server: X.Org 1.20.8 driver: modesetting,nvidia 
           unloaded: fbdev,nouveau,vesa resolution: 1920x1080~60Hz 
           OpenGL: renderer: Mesa Intel UHD Graphics 620 (KBL GT2) v: 4.6 Mesa 20.0.8 
Audio:     Device-1: Intel Sunrise Point-LP HD Audio driver: snd_hda_intel 
           Sound Server: ALSA v: k5.4.0-7642-generic 
Network:   Device-1: Intel Dual Band Wireless-AC 3168NGW [Stone Peak] driver: iwlwifi 
           IF: wlo1 state: up mac: <filter> 
           Device-2: Realtek RTL8111/8168/8411 PCI Express Gigabit Ethernet driver: r8169 
           IF: eno1 state: up speed: 1000 Mbps duplex: full mac: <filter> 
           IF-ID-1: virbr0 state: down mac: <filter> 
           IF-ID-2: virbr0-nic state: down mac: <filter> 
Drives:    Local Storage: total: 1.03 TiB used: 618.86 GiB (58.9%) 
[...]

Issue #85, opened 2020-08-07
Battery appears as Charging while not


Ratatinator97:
### Error output:

--> Battery is: charging
Battery appears as "charging" while not ! My computer was not charging at all.

---

### System information:
-------------------------------------------------------------------------------

System: HP product: HP EliteBook 830 G5 v: N/A serial: <filter> 
           Mobo: HP model: 83B3 v: KBC Version 04.65.00 serial: <filter> UEFI: HP 
           v: Q78 Ver. 01.10.01 date: 03/02/2020 
Battery:   ID-1: BAT0 charge: 20.7 Wh condition: 47.3/47.3 Wh (100%) 
CPU:       Governors: performance powersave Topology: Dual Core model: Intel Core i5-7300U bits: 64 type: MT MCP 
           L2 cache: 3072 KiB 
           Speed: 1611 MHz min/max: 400/2600 MHz Core speeds (MHz): 1: 1751 2: 1895 3: 1857 
           4: 1802 
Graphics:  Device-1: Intel HD Graphics 620 driver: i915 v: kernel 
           Device-2: Lite-On HP HD Camera type: USB driver: uvcvideo 
           Display: server: X.Org 1.20.8 driver: intel unloaded: modesetting,vesa 
           resolution: 1920x1080~60Hz 
           OpenGL: renderer: Mesa Intel HD Graphics 620 (KBL GT2) v: 4.6 Mesa 20.1.4 
Audio:     Device-1: Intel Sunrise Point-LP HD Audio driver: snd_hda_intel 
           Sound Server: ALSA v: k5.7.12-arch1-1 
Network:   Device-1: Intel Ethernet I219-LM driver: e1000e 
           IF: enp0s31f6 state: down mac: <filter> 
           Device-2: Intel Wireless 8265 / 8275 driver: iwlwifi 
           IF: wlp1s0 state: up mac: <filter> 
Drives:    Local Storage: total: 238.47 GiB used: 110.49 GiB (46.3%) 
           ID-1: /dev/sda vendor: Micron model: MTFDDAV256TBN-1AR15ABHA size: 238.47 GiB 
Partition: ID-1: / size: 229.28 GiB used: 110.38 GiB (48.1%) fs: ext4 dev: /dev/dm-2 
           ID-2: /boot size: 511.0 MiB used: 109.7 MiB (21.5%) fs: vfat dev: /dev/sda1 
[...]

PR #124, opened 2020-09-25
Rewrote charging check function

Haptein:
Adresses #85  and #108.

The solution I opted for was reading from AC-adapters and laptop batteries in /sys/class/power_supply/. With some configurations this can become a tad messy however (altho there's no other way to do this automatically afaik). Psutil actually reads from this directory, but it doesn't take into account some scenarios.

Looking into it I've seen the following kind of devices in this directory:

**AC Adapters**
- ACAD
- AC
- AC0
- ADP0
- ADP1

**Laptop Batteries**
- BAT0
- BAT1

**Peripheral Batteries** (mice, etc)
- hidpp_battery
- hidpp_battery_2
- battery_hid_somethingsomethingsomething_battery

There is also the case where a laptop has two ac-adapters and/or two batteries, and the case where the kernel can't properly read battery and/or ac-adapter information due to firmware issues. Which while investigating, turned out to be the case for my own laptop:

dmesg:
battery: ACPI: Battery Slot [BAT1] (battery present)
battery: [Firmware Bug]: battery: (dis)charge rate invalid.

Resulting in battery status 'Unknown' (when battery fully charged), and also psutil wrongly deciding it's charging state.

In the "weird" case that both ac-adapter and battery status are unknown by the kernel the function defaults to returning False which I thought appropriate (erring on the safe side). Also opted for using grep's output, because it looked too messy doing it with python.

I'd like your opinion on these liberties I've taken. Also I'd like some verification on other machines.

### DIFF
--- a/source/core.py
+++ b/source/core.py
@@ -107,13 +107,27 @@ def charging():
     """
     get charge state: is battery charging or discharging
     """
-    bat_info = psutil.sensors_battery()
-    if bat_info is None:
-        state = True
-    else:
-        state = bat_info.power_plugged
+    power_dir = "/sys/class/power_supply/"
 
-    return state
+    # AC adapter states: 0, 1, unknown
+    ac_info = getoutput(f"grep . {power_dir}A*/online").splitlines()
+    # if there's one ac-adapter on-line, ac_state is True
+    ac_state = any(['1' in ac.split(':')[-1] for ac in ac_info])
+
+    # Possible values: Charging, Discharging, Unknown
+    battery_info = getoutput(f"grep . {power_dir}BAT*/status")
+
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
I've tested the changes and they work fine for me. Code wise, also fine with "liberties" you've taken :)

I've also pushed changes to snap "beta" channel, let's see if it does anything for reports in #85 and #108 after which I can merge it with master.

AdnanHodzic:
It's been verified that this fixes the issue for those affected in #108, merging the changes with `master` and making it part of [1.3.4 release](https://github.com/AdnanHodzic/auto-cpufreq/releases/tag/v1.3.4). Thanks for you contribution!

PR #131, opened 2020-11-22
Fixed bug in performance boost logic.

dfuehrer:
Now boosts when max core usage is above 75% rather than max core usage close to 75%.

### DIFF
--- a/source/core.py
+++ b/source/core.py
@@ -372,7 +372,7 @@ def set_performance():
     # cpu usage/system load
     display_load()
 
-    if psutil.cpu_percent(percpu=False, interval=0.01) >= 20.0 or isclose(max(psutil.cpu_percent(percpu=True, interval=0.01)), 75):
+    if psutil.cpu_percent(percpu=False, interval=0.01) >= 20.0 or max(psutil.cpu_percent(percpu=True, interval=0.01)) >= 75:
         print("High CPU load, setting turbo boost: on")
         turbo(True)
     elif load1m >= performance_load_threshold:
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Good point, but why didn't you make same changes for `set_powersave` function? 

Interestingly enough I've just found that if on powersave mode, and if I stress my CPU (`stress --cpu 8 --io 4 --vm 4 --timeout 45`) max frequency I get per core is only only in range: 2100-2250MHz out of max 4600MHz, i,e: 

```
CPU max frequency: 4600 MHz
CPU min frequency: 400 MHz

	 Usage  Temperature  Frequency
CPU0:	100.0%     68 °C     2123 MHz
CPU1:	100.0%     68 °C     2124 MHz
CPU2:	100.0%     66 °C     2125 MHz
CPU3:	100.0%     70 °C     2127 MHz
CPU4:	100.0%     68 °C     2129 MHz
CPU5:	100.0%     68 °C     2128 MHz
CPU6:	 99.0%     66 °C     2127 MHz
CPU7:	100.0%     70 °C     2120 MHz
```

Any idea about this?

dfuehrer:
I didn't change the powersave logic since it has it boost when the max core usage is close to 100%.  If I changed it it would make it more like boosting when the max core usage is over 95%.

I didn't notice that but it does happen on my laptop too when it is set to the shedutil governor.  My best guess is that the max cpu frequency for the power saving governors is less than the max frequency overall but i don't know why it wouldn't update based on the governor used.  I'm not very familiar with this stuff.

AdnanHodzic:
> I didn't change the powersave logic since it has it boost when the max core usage is close to 100%. If I changed it it would make it more like boosting when the max core usage is over 95%.

Make sense.

> I didn't notice that but it does happen on my laptop too when it is set to the shedutil governor. My best guess is that the max cpu frequency for the power saving governors is less than the max frequency overall but i don't know why it wouldn't update based on the governor used. I'm not very familiar with this stuff.

Agreed, I also think this is the reason. I'll investigate this further when I have time. 

PR #134, opened 2020-12-06
Adds OpenSUSE Tumbleweed to detected distributions

NJOYSoftware:
Hi all,
I added the detection of the OpenSUSE Tumbleweed distribution to the installer script as I'm using a laptop with it and I wanted to install auto-cpufreq.

I used this commit to install it, it shouldn't give problems to others using the same distro.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -100,6 +100,21 @@ then
     separator
     complete_msg
     separator
+# OpenSUSE Tumbleweed
+elif [ -f /etc/os-release ] && eval "$(cat /etc/os-release)" && [ $ID == "opensuse-tumbleweed" ];
+then
+    separator
+    echo -e "\nDetected OpenSUSE Tumbleweed distribution\n"
+    echo -e "\nSetting up Python environment\n"
+    zypper install python38 python3-pip python3-setuptools
+    echo -e "\nInstalling necessary Python packages\n"
+    pip_pkg_install
+    separator
+    echo -e "\ninstalling auto-cpufreq tool\n"
+    install
+    separator
+    complete_msg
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Hi @NJOYSoftware, first of all thank you for your contribution.

Before I merge this with master. Why did you limit it particularly match only "opensuse-tumbleweed? Wouldn't it be better if it worked with any OpenSuse version?

If:

```
[ $ID == "opensuse-tumbleweed" ];
```
matching was more like:
```
[ $ID == "opensuse*" ];
```

NJOYSoftware:
> but could you please consider extending this condition match beyond just "opensuse-tumbleweed? Wouldn't it be better if it worked with any OpenSuse version?

I would say so, but I don't know if any opensuse distro would install the software in this way, so I didn't want to commit something that I don't know it works smoothly. In the next days, I'll setup a VM with opensuse leap to check whether it works the same or not, and I can make a new PR.

Thank you for merging my commit!

AdnanHodzic:
> > but could you please consider extending this condition match beyond just "opensuse-tumbleweed? Wouldn't it be better if it worked with any OpenSuse version?
> 
> I would say so, but I don't know if any opensuse distro would install the software in this way, so I didn't want to commit something that I don't know it works smoothly. In the next days, I'll setup a VM with opensuse leap to check whether it works the same or not, and I can make a new PR.

Good thinking, sure take your time and I'll be looking forward to any future PR you may have. 

> Thank you for merging my commit!

Of course, changes are already live as part of [v1.5.0 release](https://github.com/AdnanHodzic/auto-cpufreq/releases/tag/v1.5.0) :)

PR #135, opened 2020-12-07
Improves opensuse detection

NJOYSoftware:
Hi, I managed to create a VM and install opensuse leap, so I made this changes to install auto-cpufreq without errors:
* Adds detection of openSUSE Leap by using a regexp when checking for the distribution ID.
* Adds `python3-devel` and `gcc` packages to be installed with zypper, as a new OS Leap install does not include them, they are also added to tumbleweed to be sure they are present (and for symmetry, too).
* Changes the bash comment to include both Tumbleweed and Leap version of openSUSE.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -101,12 +101,17 @@ then
     complete_msg
     separator
 # OpenSUSE Tumbleweed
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Great stuff! Thank you for your contribution!

PR #70, opened 2020-08-05
implemented --debug option

Red-Eyed:
- Added --debug option that shows sys info
- Added inxi as dependency
- Implemented get_sys_info() that returns inxi output
  with injected governnor information
- Fixed bug: delete_file was called but this was deleted in prev PR
- Updated issue template

PR #148, opened 2021-01-19
Improve logging & prevent log size growth 

Technical27:
Now auto-cpufreq will write directly to the logfile (if running as a daemon) or stdout.
The file (instead of just stdout) will automatically be cleared when on each refresh.

I am not that experienced with python and did test this for a decent while.

### DIFF
--- a/bin/auto-cpufreq
+++ b/bin/auto-cpufreq
@@ -31,6 +31,7 @@ def main(monitor, live, daemon, install, log, debug):
     else:
         # Important: order does matter
         if daemon:
+            file_logging()
             if os.getenv("PKG_MARKER") == "SNAP" and dcheck == "enabled":
                 while True:
                     root_check()
@@ -41,7 +42,6 @@ def main(monitor, live, daemon, install, log, debug):
                     sysinfo()
                     set_autofreq()
                     countdown(5)
-                    run("clear")
             elif os.getenv("PKG_MARKER") != "SNAP":
                 while True:
                     root_check()
@@ -52,7 +52,6 @@ def main(monitor, live, daemon, install, log, debug):
                     sysinfo()
                     set_autofreq()
                     countdown(5)
-                    run("clear")
             else:
                 daemon_not_found()
         elif monitor:
@@ -66,7 +65,6 @@ def main(monitor, live, daemon, install, log, debug):
                 sysinfo()
                 mon_autofreq()
                 countdown(5)
-                run("clear")
         elif live:
             while True:
                 root_check()
@@ -78,7 +76,6 @@ def main(monitor, live, daemon, install, log, debug):
                 sysinfo()
                 set_autofreq()
                 countdown(5)
-                run("clear")
         elif log:
             read_log()
         elif debug:
--- a/scripts/auto-cpufreq.service
+++ b/scripts/auto-cpufreq.service
@@ -1,12 +1,10 @@
 [Unit]
 Description=auto-cpufreq - Automatic CPU speed & power optimizer for Linux
 After=network.target network-online.target
-ConditionPathExists=/var/log/auto-cpufreq.log
 
 [Service]
 Type=simple
 User=root
 ExecStart=/usr/local/bin/auto-cpufreq --daemon
-StandardOutput=append:/var/log/auto-cpufreq.log
 [Install]
 WantedBy=multi-user.target
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -12,7 +12,7 @@ import warnings
 from math import isclose
 from pathlib import Path
 from pprint import pformat
-from subprocess import getoutput, call, run, check_output
+from subprocess import getoutput, call, run, check_output, DEVNULL
 
 import psutil
 import distro
@@ -30,7 +30,7 @@ SCRIPTS_DIR = Path("/usr/local/share/auto-cpufreq/scripts/")
 ALL_GOVERNORS = ("performance", "ondemand", "conservative", "schedutil", "userspace", "powersave")
 CPUS = os.cpu_count()
 
-# Note: 
+# Note:
 # "load1m" & "cpuload" can't be global vars and to in order to show correct data must be
 # decraled where their execution takes place
 
@@ -38,12 +38,22 @@ CPUS = os.cpu_count()
 powersave_load_threshold = (75*CPUS)/100
 performance_load_threshold = (50*CPUS)/100
 
-# auto-cpufreq log file
-auto_cpufreq_log_file = Path("/var/log/auto-cpufreq.log")
-auto_cpufreq_log_file_snap = Path("/var/snap/auto-cpufreq/current/auto-cpufreq.log")
+# auto-cpufreq log file path
+auto_cpufreq_log_path = None
+auto_cpufreq_log_file = None
+
+if os.getenv("PKG_MARKER") == "SNAP":
+    auto_cpufreq_log_path = Path("/var/snap/auto-cpufreq/current/auto-cpufreq.log")
+else:
+    auto_cpufreq_log_path = Path("/var/log/auto-cpufreq.log")
 
 # daemon check
 dcheck = getoutput("snapctl get daemon")
+
+def file_logging():
+    global auto_cpufreq_log_file
+    auto_cpufreq_log_file = open(auto_cpufreq_log_path, "w")
+    sys.stdout = auto_cpufreq_log_file
 
 # ToDo: read version from snap/snapcraft.yaml and write to $SNAP/version for use with snap installs
 # also come up with same kind of solution for AUR
@@ -95,7 +105,7 @@ def turbo(value: bool = None):
 
  # display current state of turbo
 def get_turbo():
-   
+
     if turbo():
         print("Currently turbo boost is: on")
     else:
@@ -220,7 +230,7 @@ def deploy_daemon():
     except:
         print("\nERROR:\nWas unable to turn off bluetooth on boot")
 
-    auto_cpufreq_log_file.touch(exist_ok=True)
+    auto_cpufreq_log_path.touch(exist_ok=True)
 
     print("\n* Deploy auto-cpufreq install script")
     shutil.copy(SCRIPTS_DIR / "auto-cpufreq-install.sh", "/usr/bin/auto-cpufreq-install")
@@ -238,7 +248,7 @@ def remove():
     if not os.path.exists("/usr/bin/auto-cpufreq-remove"):
         print("\nauto-cpufreq daemon is not installed.\n")
         sys.exit(1)
-        
+
     print("\n" + "-" * 21 + " Removing auto-cpufreq daemon " + "-" * 22 + "\n")
 
     print("* Turn on bluetooth on boot")
@@ -261,8 +271,11 @@ def remove():
     os.remove("/usr/bin/auto-cpufreq-remove")
 
     # delete log file
-    if auto_cpufreq_log_file.exists():
-        auto_cpufreq_log_file.unlink()
+    if auto_cpufreq_log_path.exists():
+        if auto_cpufreq_log_file is not None:
+            auto_cpufreq_log_file.close()
+
+        auto_cpufreq_log_path.unlink()
 
     # restore original cpufrectl script
     cpufreqctl_restore()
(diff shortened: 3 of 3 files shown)


scripts/auto-cpufreq.service
```diff
@@ -1,12 +1,10 @@
 [Unit]
 Description=auto-cpufreq - Automatic CPU speed & power optimizer for Linux
 After=network.target network-online.target
-ConditionPathExists=/var/log/auto-cpufreq.log
```
AdnanHodzic, line 4: Why did you remove this? Otherwise how will systemctl know about these and use it as log?

scripts/auto-cpufreq.service
```diff
@@ -1,12 +1,10 @@
...
 
 [Service]
 Type=simple
 User=root
 ExecStart=/usr/local/bin/auto-cpufreq --daemon
-StandardOutput=append:/var/log/auto-cpufreq.log
```
AdnanHodzic, line 10: Same as with comment above

bin/auto-cpufreq
```diff
@@ -41,7 +42,7 @@ def main(monitor, live, daemon, install, log, debug):
                     sysinfo()
                     set_autofreq()
                     countdown(5)
-                    run("clear")
```
AdnanHodzic, line 44: Why was this function call removed?

source/core.py
```diff
@@ -30,21 +30,30 @@
...
-# Note: 
+# Note:
 # "load1m" & "cpuload" can't be global vars and to in order to show correct data must be
 # decraled where their execution takes place
 
 # powersave/performance system load thresholds
 powersave_load_threshold = (75*CPUS)/100
 performance_load_threshold = (50*CPUS)/100
 
-# auto-cpufreq log file
```
AdnanHodzic, line 41: I like where you're going with this. But I'm not sure if it's due to these changes or some other changes. But if I install auto-cpufreq using `auto-cpufreq-installer --install` and then run `auto-cpufreq --install` I'll end up with following error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/var/log/auto-cpufreq.log'
```

AdnanHodzic (changes requested):
Some changes will be needed, while I like what was done here, current changes will break auto-cpufreq install.

bin/auto-cpufreq
Technical27: `clear_logs` automatically clears the console or logfile depending on which is used. If the log file is used it runs `truncate(0)` which clears the file, otherwise it does `run("clear")`.

scripts/auto-cpufreq.service
Technical27: My idea was to make auto-cpufreq responsible for logging directly to the file instead of systemd.

source/core.py
Technical27: Should be fixed now, It now only opens the log file if it is needed.

scripts/auto-cpufreq.service
AdnanHodzic: Right, but problem with this (or something else is causing this?) is that with these changes if I install auto-cpufreq, `auto-cpufreq --log` won't return any logs and `/var/log/auto-cpufreq.log` will be empty.

bin/auto-cpufreq
AdnanHodzic: Okay, now I see it, good point and thanks for clarification.

source/core.py
AdnanHodzic: Yes, thanks it's fixed but now the log is empty as reported in [comment above](https://github.com/AdnanHodzic/auto-cpufreq/pull/148#discussion_r561645729).

AdnanHodzic:
Considering some big changes were made as part of #150, #151, #152. #153 and I would like these changes to be part of [next 1.5.3 release](https://github.com/AdnanHodzic/auto-cpufreq/releases).  In order to avoid conflicts, could you just make sure you have latest changes as part of this PR, thanks!

scripts/auto-cpufreq.service
librewish: Actually i remove this line in garuda linux
Via a hook everytime the package is installed/updated
Because if its present the service fails on garuda linux

So i think removing it from source will 
Make me remove this workaround

Technical27: I don't think that line was every nessessary. In systemd's [man page](https://man.archlinux.org/man/systemd.exec.5.en): "If path refers to a regular file on the filesystem, it is opened (created if it doesn't exist yet)". All the `ConditionPathExists` does is to stop the service from starting when the file doesn't exist.

AdnanHodzic: @Technical27 now (579fec7) the log is not empty, but it'll write to log file only once. 

Live mode (`auto-cpufreq --live`) will freeze at `"auto-cpufreq" refresh in: 1`, i.e

```

Linux distro: Ubuntu 20.04 focal
Linux kernel: 5.4.0-60-generic
Cores: 8
Architecture: x86_64
Driver: intel_pstate

------------------------------ Current CPU states ------------------------------

CPU max frequency: 4600 MHz
CPU min frequency: 400 MHz

	 Usage  Temperature  Frequency
CPU0:	  4.0%     70 °C     3099 MHz
CPU1:	  5.9%     61 °C     4216 MHz
CPU2:	  1.0%     62 °C     4268 MHz
CPU3:	  2.0%     62 °C     4306 MHz
CPU4:	  3.0%     70 °C     3637 MHz
CPU5:	  0.0%     61 °C     4511 MHz
CPU6:	  6.1%     62 °C     4418 MHz
CPU7:	  3.0%     62 °C     4326 MHz

Battery is: charging

Total CPU usage: 2.6 %
Total system load: 1.18
Average temp. of all cores: 63.75 °C

Load optimal
Optimal total CPU usage: 2.6 %, high average core temp: 63.75 °C
setting turbo boost: off

			"auto-cpufreq" refresh in: 1
```

Install (`auto-cpufreq --install`) will go through, but once I check logs (`auto-cpufreq --log`) I'll see same results as with live mode. 

Furthermore, after I do uninstall I'll end up with following error:

```
sudo auto-cpufreq --remove

--------------------- Removing auto-cpufreq daemon ----------------------

* Turn on bluetooth on boot

------------------ Running auto-cpufreq daemon removal script ------------------

* Stopping auto-cpufreq daemon (systemd) service

* Disabling auto-cpufreq daemon (systemd) at boot
Removed /etc/systemd/system/multi-user.target.wants/auto-cpufreq.service.

* Removing auto-cpufreq daemon (systemd) unit file

* Reloading systemd manager configuration
reset failed
  File "/usr/local/bin/auto-cpufreq", line 4, in <module>
    __import__('pkg_resources').run_script('auto-cpufreq==1.0', 'auto-cpufreq')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 667, in run_script
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1470, in run_script
  File "/usr/local/lib/python3.8/dist-packages/auto_cpufreq-1.0-py3.8.egg/EGG-INFO/scripts/auto-cpufreq", line 138, in <module>
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 829, in __call__
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1066, in invoke
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 610, in invoke
  File "/usr/local/lib/python3.8/dist-packages/auto_cpufreq-1.0-py3.8.egg/EGG-INFO/scripts/auto-cpufreq", line 133, in main
  File "/usr/local/lib/python3.8/dist-packages/auto_cpufreq-1.0-py3.8.egg/auto_cpufreq/core.py", line 273, in remove
NameError: name 'auto_cpufreq_log_file' is not defined
```

Technical27: Those should be fixed now, and it clears the log file when the daemon starts. I don't know about live mode, but I think that is also fixed because it works for me without freezing.

AdnanHodzic: Thanks @Technical27! With cb9fcfa everything is working great. I just have one last question before I merge this with `master`.  I've installed auto-cpufreq and I've been observing its log size (`/var/log/auto-cpufreq.log`) which seems to be constantly growing. In some 15mins of my observations it grow to 124k (it grows at rate 4k every 20-30 seconds). 

As IIRC initial goal of this PR was to stop growing log size..

Technical27: I compromised on clearing the log file on startup, so that you could still get logs, but the log wouldn't grow as much as before. My problem was that nothing would clear the log file other than just manually clearing it. Maybe clearing the log after 10/20 countdowns would be a better idea along with clearing on startup.

AdnanHodzic: > I compromised on clearing the log file on startup, so that you could still get logs, but the log wouldn't grow as much as before.

I'm not sure about only clearing it on startup, as for example I rarely reset my laptop. I'm still tracking size of the log, and now some 30 minutes in it's at 580K

> My problem was that nothing would clear the log file other than just manually clearing it.

Instead of doing what you're doing in `file_logging` function now, with `truncate` you could simply do:

```
open("/var/log/auto-cpufreq.log", "w").close()
```

It's bit of a hack, but it'll do the job and empty the file. 

> Maybe clearing the log after 10/20 countdowns would be a better idea along with clearing on startup.

I agree, I would even clear the log at every countdown, as new data is written to it regardless.

Technical27: My thought was it has to open the file anyway, so write directly to the file instead of relying on systemd to do that for us. I think that combining clearing after every countdown and on start is a good idea.

AdnanHodzic: > I think that combining clearing after every countdown and on start is a good idea.

Sounds good, let me know once you've made these changes as it's bit hard to keep track with force pushes. Also, consider using `open("/var/log/auto-cpufreq.log", "w").close()` to empty the file.

AdnanHodzic: > I think that combining clearing after every countdown and on start is a good idea.

Sounds good, let me know once you've made these changes as it's bit hard to keep track with force pushes. With last commit (de79f01), log size is still growing.  

 As I suggested, consider using `open("/var/log/auto-cpufreq.log", "w").close()` to empty the file.

Technical27: It should work now on ace152a.

Technical27: @AdnanHodzic  I tested it, It is actually fixed in b3ed601.

AdnanHodzic: @Technical27, nice, looks good. There's one minor thing, when looking at the log and when countdown hits 0, following message will appear: `"auto-cpufreq" refresh in: 1tail: /var/log/auto-cpufreq.log: file truncated`

It would be nice to get rid of `tail: /var/log/auto-cpufreq.log: file truncated` part.

Technical27: Fixed in 6c15cae. Though there is the downside that tail won't print any errors, but the alternatives would be more complicated.

AdnanHodzic: Hm, I'm not sure about this approach and not printing errors. As a lot of raised issues are by people seeing some error in the log file. 

The way I would approach this is if string in output starts with "tail:" omit it and whatever's printed after it. I remember there was quite some output being omitted as part of [original "debug" functionality](https://github.com/AdnanHodzic/auto-cpufreq/pull/70/files) using Python re ...

Also, it's clear that once it gets to  `"auto-cpufreq" refresh in: 1` it stays there for more then 1 seconds, but that's a minor problem and is negligible.

Technical27: I mean that if `tail` itself has an error, it doesn't print it, but any error from auto-cpufreq will be printed to the log file and `tail` will show that.

Technical27: I man that `tail` won't print any errors, but other errors will show up. So in the unlikely event that `tail` screws up, that won't print any errors.

AdnanHodzic: You're right. Let's get this merged, thank you for your contribution and tenacity in this pull request :)

AdnanHodzic (approved):

PR #152, opened 2021-01-20
Add Arch-based Linux to auto-cpufreq-installer script.

marc0der:
Adds Arch Linux (and Manjaro) to installer script using `pacman` to install dependencies.


AdnanHodzic:
Great stuff, thank you for your contribution!

PR #158, opened 2021-01-31
fix problem in desktops where there is nothing in /sys/class/power_su…

### LINKED ISSUE #127: CPU frequency doesn't automatically adjust
No error output, but there seems to be a bug where the CPU doesn't scale up its' frequency when doing tasks such as gaming. I posted about the issue on https://forum.garudalinux.org/t/csgo-microstutters/776 . I'm sorry if this isn't the right way to post about the issue, but I am new to Linux and this is my first post on GitHub.

### Error output:
```text
● auto-cpufreq.service - auto-cpufreq - Automatic CPU speed & power optimizer for Linux
     Loaded: loaded (/usr/lib/systemd/system/auto-cpufreq.service; enabled; vendor preset: disabled)
     Active: active (running) since Mon 2020-10-05 17:34:45 CEST; 2h 5min ago
   Main PID: 2106 (auto-cpufreq)
      Tasks: 2 (limit: 19060)
     Memory: 31.3M
     CGroup: /system.slice/auto-cpufreq.service
             ├─  2106 /usr/sbin/python /usr/bin/auto-cpufreq --daemon
             └─240652 /bin/sh -c cpufreqctl --driver

okt. 05 17:34:45 linux-desktop systemd[1]: Started auto-cpufreq - Automatic CPU speed & power optimizer for Linux.
```

librewish:
fixes #127
on some desktop pc there is nothing in /sys/class/power_supply

and this caused desktops to in powersave

desktops should be on performance

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -117,10 +117,16 @@ def charging():
     """
     power_dir = "/sys/class/power_supply/"
 
-    # AC adapter states: 0, 1, unknown
-    ac_info = getoutput(f"grep . {power_dir}A*/online").splitlines()
-    # if there's one ac-adapter on-line, ac_state is True
-    ac_state = any(['1' in ac.split(':')[-1] for ac in ac_info])
+    ac_online = Path("/sys/class/power_supply/A*/online")
+    
+    if ac_online.exists():
+        # AC adapter states: 0, 1, unknown
+        ac_info = getoutput(f"grep . {power_dir}A*/online").splitlines()
+        # if there's one ac-adapter on-line, ac_state is True
+        ac_state = any(['1' in ac.split(':')[-1] for ac in ac_info])
+    else:
+        # if there is no ac_online return ac_sate as true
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Unfortunately while this might've fixed problems on desktops. With these changes my laptop (Lenovo X1 Carbon) is constantly reporting that `Battery is: charging` whether this is true or not. 

There are couple of ways you could do this, you first want to make check if you're on laptop or desktop. Easiest way to do this is by running: `hostnamectl status | grep Chassis | cut -f2 -d ":" | tr -d ' '`

If laptop it'll return `laptop` otherwise it should be `desktop`.

Alternatively, see what's done as part of ["charging" function](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto_cpufreq/core.py#L114) and potentially utilize what's being done in `ac_state` or `battery_info` to make similiar kind of check.

AdnanHodzic:
PR continued in #162

PR #160, opened 2021-01-31
Improve domain language around logging and stats

marc0der:
This PR attempts to improve the logging/stats domain names as per #156:

* renames all stdout occurrences of `log` to `stats`
* update all code references accordingly
* rename `.log` file to `.stats`
* move the stats file from `/var/log` to `/var/run`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,6 @@ dmypy.json
 
 # snap
 *.snap
+
+# pycharm
+.idea/
--- a/README.md
+++ b/README.md
@@ -95,11 +95,11 @@ Necessary changes are temporarily made to the system which are lost with system 
 
 ### Install - auto-cpufreq daemon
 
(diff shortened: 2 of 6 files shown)


AdnanHodzic:
Code changes LGTM. 

As I proposed in #156, only change I'd propose/request is to  leave `--log` hidden flag there for awhile, if whoever runs `sudo auto-cpufreq --log` that they should get a message "Logs option is no longer in use, please use `sudo auto-cpufreq --stats` instead." or something along those lines.

marc0der:
@AdnanHodzic that's done and tested :+1:

AdnanHodzic:
@marc0der great stuff! I think it would be a nice idea to add a "time stamp" at the end of each cycle in `--stats` for better reference when scrolling through these, but for now it's perfect as it is.

Thank you for you contribution and I look forward to any new ideas and/or suggestions you may have in future :)

PR #162, opened 2021-02-02
Adjust CPU frequency scaling on desktop to run in "performance" governor

ThePoorPilot:
This is an updated version of the commit submitted by @librewish | PR: #158  (Fixes #127)

If it detects the device is a laptop, it should default to the regular method of determining ac_state
If the device returns any other output, it should default to ac_state=true

This is my first commit on github ever. I still don't really know much about programming, but with some research, I figured it out.
Couldn't fully test it because I had difficulty running it on arch(Garuda Linux), but I think this should work in theory.

Please test it out on your thinkpad or use your more coding-oriented brain to weed out obvious problems. I got it to compile with no errors and I could run the command "auto-cpufreq", but couldn't get it to run as a system daemon.

Also, sorry if this pull request is not done exactly properly, I am not very familiar with github

PR #163, opened 2021-02-03
use dmidecode and dependancies update

librewish:
Continuation/links to: #162


AdnanHodzic:
Great stuff, thank you for your contribution!

ThePoorPilot:
Ah, it looks like hostnamectl is a part of systemd. Although most people use systemd, there is a sizable portion of people who don't, this is a good catch. Otherwise, this package would require systemd to operate. I just pulled the command from Adnan's comment on pull request #158 .

Issue #164, opened 2021-02-03
System is discharging but auto-cpufreq detects charging


pheiduck:
### Error output:
```
Show charging but it's discharging
```
---

### System information:
```
-------------------------------------------------------------------------------

Linux distro: Manjaro Linux 20.2.1 Nibia
Linux kernel: 5.10.12-3-MANJARO
Processor: Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
Cores: 4
Architecture: x86_64
Driver: intel_cpufreq

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 2900 MHz
CPU min frequency: 500 MHz

Core	Usage	Temperature	Frequency
CPU0:	  0.0%     56 °C     2415 MHz
CPU1:	  4.0%     56 °C     2259 MHz
CPU2:	  2.0%     56 °C     2124 MHz
CPU3:	  1.0%     56 °C     1800 MHz

fatal: Kein Git-Repository (oder irgendein Elternverzeichnis bis zum Einhängepunkt /)
Stoppe bei Dateisystemgrenze (GIT_DISCOVERY_ACROSS_FILESYSTEM nicht gesetzt).
Snap package: no

Python: 3.9.1
psutil package: 5.8.0
platform package: 1.0.8
click package: 7.1.2
distro package 1.5.0

Battery is: charging

auto-cpufreq system resource consumption:
cpu usage: 0.0 %
memory use: 0.21 %

Total CPU usage: 2.0 %
Total system load: 0.68
Average temp. of all cores: 56.0 °C 

Currently using: performance governor
Currently turbo boost is: on

------------------------------------------------------------------------------
```
---

ThePoorPilot:
First, what version of auto-cpufreq are you running?
The recent update fixed an issue for desktops, but it could come at the detriment of laptops.

If you are running the latest version, what is the output of the command "hostnamectl status | grep Chassis | cut -f2 -d ":" | tr -d ' '

AdnanHodzic:
@pheiduck make sure you're using the latest version, [1.5.4 was released](https://github.com/AdnanHodzic/auto-cpufreq/releases/tag/v1.5.4) few hours ago. If AUR package is older, you can also install latest version using [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq#auto-cpufreq-installer).

Could you also return output of `dmidecode --string chassis-type` and please confirm that you're running on a laptop.

[...]

PR #167, opened 2021-02-03
WIP: More extensive chassis detection for laptops

ThePoorPilot:
I think this should work. This should address issue #164 

Once again, I don't have a way to fully test it. I tried compiling from the AUR and replacing the github link in the PKGBUILD to my fork, but I couldn't get it to work. I don't have access to the pkgbuild used on the Chaotic-AUR, unfortunately, because that would likely work.

Anyways, this sets the SMBIOS Chassis types that don't have batteries to ac_state=true
This switches the if else order from pull request #163.
In the new order, the default state if a desktop is not detected is a laptop.

I also added computer type to the  auto-cpufreq --stats command. It is up to your discretion whether or not that is a worthwhile addition. I think it could be useful, but it is not necessary. 

Update to core.py fixes variable not found for computer_type in the --stats
I managed to get it to kind of run, but auto-cpufreq --stats doesn't work. Only sudo auto-cpufreq --monitor works, but that's enough to see that the issue is fixed

Links to: #166, #164

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -119,14 +119,62 @@ def charging():
 
     computer_type = getoutput('dmidecode --string chassis-type')
     
-    if computer_type == "Notebook":
+    if computer_type == "Desktop":
+        ac_state = True
+    elif computer_type == "Low Profile Desktop":
+        ac_state = True
+    elif computer_type == "Tower":
+        ac_state = True
+    elif computer_type == "Mini Tower":
+        ac_state = True
+    elif computer_type == "All in One":
+        ac_state = True
+    elif computer_type == "Pizza Box":
+        ac_state = True
+    elif computer_type == "Lunch Box":
+        ac_state = True
+    elif computer_type == "Main Server Chassis":
+        ac_state = True
+    elif computer_type == "RAID Chassis":
+        ac_state = True
+    elif computer_type == "Rack Mount Chassis":
+        ac_state = True
+    elif computer_type == "Blade":
+        ac_state = True
+    elif computer_type == "Blade Enclosure":
+        ac_state = True
+    elif computer_type == "Multi-system chassis":
+        ac_state = True
+    elif computer_type == "Sealed-case PC":
+        ac_state = True
+    elif computer_type == "Mini PC":
+        ac_state = True
+    elif computer_type == "Stick PC":
+        ac_state = True
+    elif computer_type == "Embedded PC":
+        ac_state = True
+    elif computer_type == "Advanced TCA":
+        ac_state = True
+    elif computer_type == "Compact PCI":
+        ac_state = True
+    elif computer_type == "IoT Gateway":
+        ac_state = True
+    elif computer_type == "Expansion Chassis":
+        ac_state = True
+    elif computer_type == "Bus Expansion Chassis":
+        ac_state = True
+    elif computer_type == "SubChassis":
+        ac_state = True
+    elif computer_type == "Peripheral Chassis":
+        ac_state = True
+    elif computer_type == "Space-saving":
+        ac_state = True
+    else:
+        #Defaults to laptop power determination
         # AC adapter states: 0, 1, unknown
         ac_info = getoutput(f"grep . {power_dir}A*/online").splitlines()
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
No, I will not accept it like this. If we are going to go this way (and we're not) this should be placed in a map/dictionary/list. Furthermore, let's please focus that this tool is primarily made to be running on portable devices with battery (laptops). So we shouldn't have cases where it's running on "Pizza Box" (whatever that is). That should go in "other/else" case. 

Changes that were introduced here broke one of the core auto-cpufreq functionalities, which I take very seriously. In my lack of time, I was very tempted to revert all changes, but instead to mitigate this problem I just made[ simple changes](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto_cpufreq/core.py#L122) and [released "hot-fix"](https://github.com/AdnanHodzic/auto-cpufreq/releases/tag/v1.5.5) that should encompass "laptop" cases which I got by testing this on 3 different (Intel) laptops. 

**So let's please focus on finding all different cases there are of what `dmidecode --string chassis-type` when it's run on different laptops and extend this list.** I'm particularly interested what will be returned for AMD or ARM laptops. On this topic @Red-Eyed could you please tell me what you get for mentioned dmidecode output? Thank you.

> I also added computer type to the auto-cpufreq --stats command. It is up to your discretion whether or not that is a worthwhile addition. I think it could be useful, but it is not necessary.

I added this to `--debug` as it could only be useful there.

librewish:
https://morecomputerstuff.net/2017/06/27/smbios-chassis-type-table/

here there are other cases in which its laptop

chasis types

Portable
Laptop
Notebook
Hand Held
Sub Notebook
Tablet
Convertible
Detachable

librewish:
also here
https://gist.github.com/Spartan-196/e637a67ede6e1c99f0607feceb2de87c

AdnanHodzic:
Thanks @librewish! Can one of you 2 start a PR to extend this list? We should give it some time to hear what everybody says as [part of my Tweet](https://twitter.com/fooctrl/status/1357230198857564160). As I said, there are also interesting [edge cases](https://twitter.com/LokiC3b/status/1357234945765367809) which should also be dealt with, and there might be more.

librewish:
how about detecting battery also in the else part
if its present 
then do the charging detection
else
set charging = true

librewish:
@AdnanHodzic 
what do you think of using psutil
https://stackoverflow.com/a/41988627

librewish:
has_battery = psutil.sensors_battery() is not None
    
    if has_battery == True:
        power_pluggedin = psutil.sensors_battery().power_plugged
        if power_pluggedin == True:
            ac_state = True
        else:
            ac_state = False
    else:
        ac_state = True

ThePoorPilot:
It sounds like this is getting more complicated than I can really deal with.

I do want to apologies apologize for how haphazardly the first solution was put together.

 Chassis type is not as a perfect metric it seemed

The reason I did it to check for desktops, was because I figured everything else would fall under the "else" category. I did consider add elif statements for "Notebook" and "Laptop," but I considered it redundant.

Also I apologize for it being such a long list of elifs, I just figured it was more important than to have something than for me to give up trying to figure out how to check for multiple conditions with an if statement.

As Librewish says, maybe something close to his initial approach, but detecting for battery could be more optimal. We'd just have to get that to actually work.

I haven't read about any other projects that try to detect something like this, so we're kind of on our own to figure it out.

AdnanHodzic:
> This may be getting more complicated than I initially thought.
> 
> I do want to apologize for how haphazardly the first solution was put together.

@ThePoorPilot 

Hey, don't sweat it! It's hard to predict all use cases since there's so many of them. I also overlooked some things, these things happen, but what we do with this information is what matters.

> ```
> has_battery = psutil.sensors_battery() is not None
>     
>     if has_battery == True:
>         power_pluggedin = psutil.sensors_battery().power_plugged
>         if power_pluggedin == True:
>             ac_state = True
>         else:
>             ac_state = False
>     else:
>         ac_state = True
> ```

@librewish 

I like it! Let's close this MR and continue this in #168

PR #173, opened 2021-02-07
Only query pacman when on arch and not a snap install.

marc0der:
This is a minor fix that prevents calling out to pacman when the package is installed with snap on arch linux.

It is a minor improvement on #171.


AdnanHodzic:
Awesome, must've missed it during testing.

PR #187, opened 2021-03-09
Fix location of auto-cpufreq in service descriptor.

marc0der:
This fixes a bug that causes the systemd service to go into a failed state because the `auto-cpufreq` script can't be found.

### DIFF
--- a/scripts/auto-cpufreq.service
+++ b/scripts/auto-cpufreq.service
@@ -5,6 +5,6 @@ After=network.target network-online.target
 [Service]
 Type=simple
 User=root
-ExecStart=/usr/local/bin/auto-cpufreq --daemon
+ExecStart=/usr/bin/auto-cpufreq --daemon
 [Install]
 WantedBy=multi-user.target


marc0der:
I just installed it again from the `auto-cpufreq-installer` to make sure and it works perfectly for me.

Regarding the weirdness that you saw: the stats command initially does hang for 1 second but then renders fine without issues.

AdnanHodzic:
> I just installed it again from the `auto-cpufreq-installer` to make sure and it works perfectly for me.
> 
> Regarding the weirdness that you saw: the stats command initially does hang for 1 second but then renders fine without issues.

Yea, it's just weird. Tried it on another laptop and no problems whatsoever.

Either way, thank you for your contribution!

theol0403:
On Fedora 34, I had to revert this commit for the tool to work. Otherwise, the stats command did not print any information.

```
which auto-cpufreq
/usr/local/bin/auto-cpufreq
```

I'm not sure if I'm missing something obvious here but running the installer script, then installing the service, leads to this non-working state.

Thanks for the great tool btw!

AdnanHodzic:
> On Fedora 34, I had to revert this commit for the tool to work. Otherwise, the stats command did not print any information.
> 
> ```
> which auto-cpufreq
> /usr/local/bin/auto-cpufreq
> ```
> 
> I'm not sure if I'm missing something obvious here but running the installer script, then installing the service, leads to this non-working state.
> 
> Thanks for the great tool btw!

Thank you for your input @theol0403!

@marc0der this confirms problem I was experiencing, please let me know if you have any additional input on this or how to resolve this issue, otherwise I'll have to revert this change.

marc0der:
I've had a further look, and it seems like there are some discrepancies in the distros as to where the files are copied to. The fix I did seems to work on Arch but breaks on other distros like Fedora.

I've now [raised a new PR](https://github.com/AdnanHodzic/auto-cpufreq/pull/190) that reverts the change to the systemd unit file and customises it for Arch in the installer script.

AdnanHodzic:
@marc0der thanks! It works for me. Could @theol0403 you confirm #187 works for you as well? After which I'll have it merged with `master`

theol0403:
Seems to work for me. Many thanks!

PR #190, opened 2021-03-14
Fix service unit file on Arch Linux

marc0der:
Fixes a bug introduced by https://github.com/AdnanHodzic/auto-cpufreq/pull/187

Changes the systemd unit file inplace when installing on Arch from the installer script.


AdnanHodzic:
This is great, as always thank you for your contribution!

PR #206, opened 2021-06-01
Adding Support for void linux

shubham-cpp:
The packages are very similar to that of arch so I dont think this is create a much of a difference


shubham-cpp:
Sorry, ignore this pull request, void linux uses runit has init system so will have to further check if everything is working as expected

PR #211, opened 2021-06-24
added endeavouros for arch based system

adam4nj:
added endeavour os as a target for arch based systems


AdnanHodzic:
Sorry for such delayed reply/review.

LGTM and thank you for your contribution!

PR #235, opened 2021-08-20
Added Void Linux support (runit)

jpnt:


AdnanHodzic:
@jpnt looks great, thank you for you contribution!

Issue #200, opened 2021-05-31
[Request] Add support for artix (non systemd support)


Juliaria08:
### Error output:
```text
Systemd is not installed, wont work
```
Consider supporting [Artix Linux](https://artixlinux.org/)

AdnanHodzic:
Related: #29

AdnanHodzic:
Support for Artix has been added as part of #237. 

Please pull the latest changes and perform install using [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer)

AdnanHodzic:
Problem fixed again as part of: #706

PR #237, opened 2021-08-27
Artix runnit support

explosion-mental:
seemed easy enough to add it thanks to void linux steps. Although artix supports other inits which I haven't tried it. Maybe checking if the `runnit` directory exist before installing the daemon, else print manual installations..
```bash
if [ -d /etc/runit ] && [ -d /run/runnit ]; then
...
```

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -128,7 +128,7 @@ function tool_install {
     separator
     echo -e "\ninstalling auto-cpufreq tool\n"
     install
-    if [[ ! $ID == "artix" ]]; then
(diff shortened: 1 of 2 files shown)


AdnanHodzic:
Changes LGTM, thank you for your contribution! 

> Although artix supports other inits which I haven't tried it. Maybe checking if the `runnit` directory exist before installing the daemon, else print manual installations..

There was a feature request to add support for Artix #200 but it wasn't specified which init to support (except non systemd). Either way, feel free to extend it to whatever degree you desire and I'll be looking forward to review your Pull Requests.

Issue #26, opened 2020-01-09
[feature request/enhancement] settings file


marc-vdm:
Would it be a good idea to have a settings file?

Certain things could be stored here like;
- minimum CPU load before going to turbo 
- minimum load factor before going to turbo
- true/false for allowing turboboost to be switched off when on battery
And any other items that can be changed.

Exorust:
This is similar to #30.

AdnanHodzic:
Duplicate of #30 but let's continue the discussion here.

As I suggested in #30 

> I like this idea. My idea (which I had in plan for one of the future releases) was to allow users to specify these themselves using CLI. Something like having another "advanced" mode which would allow you to easily/interactively tweak these to your own liking.

what you're proposing here would make perfect sense. In meantime I just want to think about what would be simpler option for the end user. Adding these interactively through a CLI or a "settings" file.

e4BdSBmUzHowFico5Ktn:
I would personally say that a file gives a good overview to users of what settings there are and how they could be changed, and perhaps implementation is easier, but I'm not sure of that.
 However, there are probably more opinions on this so I`d be happy to hear those.

poVoq:
Maybe at least some default presets would be good? For example one that stays low as long as possible and only switches to performance when starting a game for something like that?
Right now it seems quite trigger happy with switching up the CPU speed on regular browser use quite often.

AdnanHodzic:
> I would personally say that a file gives a good overview to users of what settings there are and how they could be changed, and perhaps implementation is easier, but I'm not sure of that.
> However, there are probably more opinions on this so I`d be happy to hear those.

Let's no go with what's _easiest_ to implement, let's go with what's _best_ to implement :)

> Maybe at least some default presets would be good? For example one that stays low as long as possible and only switches to performance when starting a game for something like that?
> Right now it seems quite trigger happy with switching up the CPU speed on regular browser use quite often.

Having various presets is also a good idea. 

I agree that in current implementation it is trigger happy. I'm thinking of different ways to make it less this way, by possibly combining both input from load and CPU usage (and then in future also CPU temperature + fan speeds).

Either way I'll be thinking what's the best way to proceed with this in upcoming period. In meantime if anyone else is interested in having this feature, I'd be very interested in hearing what you have to say.

Red-Eyed:
I could look at it. and maybe implement first draft when have a time
[...]

PR #250, opened 2021-10-04
Add config file for preferred governor and turbo boost behaviour.

zoitrok:
This branch adds support for a configuration file to allow users to set preferred governors and turbo boost behavior for when connected a power source and when running on battery.

In the case of my devices powersave is not necessarily the best cpu governor when on battery.


AdnanHodzic:
Hi @zoitrok this looks great and is a long/high requested feature! #26 

I have few suggestions and remarks before we get this merged. 

**README**

As part of this PR I would also like to have README updated, in particular to have a new "Configuration/Settings file" section which elaborates on how a config file can be used. As currently if you don't know how to read Python you won't be able to figure out how to use it :)

**Explicitly state that conf file is used**

Although in auto-cpufreq output it's stated when conf file is used, I would be even more explicit.

When auto-cpufreq is first started "Note: You can quit live mode by pressing "ctrl+c"" message is printed. It would be good idea if right below that line something like: "Using settings defined in /etc/auto-cpufreq.conf file" is stated, just for additional clarity.

**Changes to .conf file**

Then as part of auto-cpufreq.conf file:

```
# preferred governor
governor = performance
```

Maybe it would be a good idea as part of `# preferred governor` to also have a comment how to get a list of available governors, i.e: `cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors`

Or what would be an ideal is to extend `auto-cpufreq --debug` so it can also show list of available governors, and then user could refer and see what they could select. 

It also might not be a bad idea to point to new contributors who are interested on how they could extend auto-cpufreq.conf file with their field of interest, but this is optional as they should be able to look at code and figure how it works.

Either way, thank you again and looking forward to your update and/or comments!

Adnan

PR #255, opened 2021-10-17
Disable turbo when config is "never"

fpoli:
I suspect that this is a copy-paste error.


AdnanHodzic:
Great catch! Thank you for your contribution @fpoli

PR #259, opened 2021-10-17
install service file to proper location; enable and start service

aroundthfur:
The systemd unit file was not deployed to the right location.
Adding 2 new functions to deploy the unit file and to enable and start the service.

Also added cleanup for the unit file.


AdnanHodzic:
Good stuff, thank you for your contribution @aroundthfur!

PR #263, opened 2021-10-17
WIP: Implement python virtual environment 

aroundthfur:
* added venv instead of using system env pip
  * adjusted the unit file to startup the app from the venv
  * created a wrapper script to call the app from the venv
  * extended cleanup for venv and additional scripts
  * refactored the is_running() function to find the process now that it
    is called from a venv

I tested this on Fedora 35. It worked fine (TM).
Please test on other platforms before merging. I am especially not sure how this change reflects on SNAP and other packaging systems.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -35,6 +35,7 @@ function setup_venv {
     python3 -m venv "${venv_dir}"
 
     source "${venv_dir}/bin/activate"
+    python3 -m pip install --upgrade pip
     python3 -m pip install -r requirements.txt
 }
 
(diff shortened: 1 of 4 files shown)


bobslept:
Could it be that power-profiles-daemon had something to do with it at the time of testing this?

bobslept:
What would be the correct way of doing this. Merge his branch from the fork locally, work on it and push it as new PR and mention this PR in the new one? 

I can do it like that if you want. Not sure if it is the right way.

AdnanHodzic:
> What would be the correct way of doing this. Merge his branch from the fork locally, work on it and push it as new PR and mention this PR in the new one?
> 
> I can do it like that if you want. Not sure if it is the right way.

Please do this. As I also got @aroundthfur consent to do this (make new PR on top of his changes) in the end I'll just credit you both.

AdnanHodzic:

PR #269, opened 2021-10-24
Minor fixes for Void Linux

### LINKED ISSUE #265: [BUG] A few concern and errors in Void Linux.
https://github.com/AdnanHodzic/auto-cpufreq/blob/893ed43bdfc414d4116c9b51dd4e03504fad2959/auto-cpufreq-installer#L133

`-Sy` option should be enough for syncing the repositories and yes (no-confirm). That `-u` parameter buggs me, to update whole system while installing the package.

Animeshz:
Closes #265. (summary: commit msg).

Edit: `bluez` package provides bluetooth access using traditional `bluetoothctl` and installs the `/etc/bluetooth/main.conf` file, which is not present in the base installation. So did a check if it is present, and try to change if its present.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -130,7 +130,7 @@ elif [ -f /etc/os-release ];then
 		;;
 	void)
 		detected_distro "Void Linux"
-    		xbps-install -Suy python3 python3-pip python3-devel python3-setuptools base-devel dmidecode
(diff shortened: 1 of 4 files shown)


AdnanHodzic:
Hi @Animeshz, sorry for such delayed reply. Changes LGTM and thank you for your contribution!

Issue #234, opened 2021-08-19
Always reads battery state as discharging, thus stuck on "powersave" governor on Lenovo laptop


AmitGolden:
###  Problem:
I'm using auto-cpufreq on a Lenovo ideapad 330s.
My laptop is charged, and full, but auto-cpufreq stays on "powersave" governor.
I updated my BIOS recently.
My battery state:
```text
$ cat /sys/class/power_supply/BAT0/status
Full
```
---

### System information:
```text
-------------------------------------------------------------------------------

Linux distro: Arch Linux rolling n/a
Linux kernel: 5.13.10-zen1-1-zen
Processor: Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
Cores: 8
Architecture: x86_64
Driver: intel_pstate

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 1600 MHz
CPU min frequency: 400 MHz

Core	Usage	Temperature	Frequency
CPU0:	13.0%     50 °C      752 MHz
CPU1:	11.0%     49 °C      771 MHz
CPU2:	10.0%     48 °C     1800 MHz
CPU3:	12.7%     48 °C     1800 MHz
CPU4:	 6.9%     50 °C     1800 MHz
CPU5:	 9.0%     49 °C     1800 MHz
CPU6:	 6.0%     48 °C     1800 MHz
CPU7:	 9.1%     48 °C      777 MHz

auto-cpufreq version:
Version         : 1.6.4.r225.3807d6d-1

Python: 3.9.6
psutil package: 5.8.0
platform package: 1.0.8
click package: 8.0.1
distro package 1.5.0

Computer type: Notebook
Battery is: discharging

auto-cpufreq system resource consumption:
cpu usage: 0.0 %
memory use: 0.21 %

Total CPU usage: 9.8 %
Total system load: 1.74
Average temp. of all cores: 48.75 °C 

Currently using: powersave governor
Currently turbo boost is: off

-------------------------------------------------------------------------------
```
---

stepan-neretin7:
Same bug

Haptein:
grep . /sys/class/power_supply/A*/online
ls /sys/class/power_supply/

What's the output of these when that happens? I kinda already know the fix but I'm curious about what your ac adapter state is.

AmitGolden:
`$ ls /sys/class/power_supply/
BAT0`
[...]

PR #274, opened 2021-11-09
On laptops and notebooks, check if charging via BAT* status

AmitGolden:
Hi,
For some time now auto-cpufreq got stuck on the powersave governor for me.
This happens because Linux does not recognize my power adapter for some reason, thus using `grep . /sys/class/power_supply/A*/online` doesn't work for me.

I discussed this further in #234.

So, I changed `charging()` to read from `/sys/class/power_supply/BAT*/status` instead, which will greatly improve my experience.

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -168,10 +168,10 @@ def charging():
 
     computer_type = getoutput("dmidecode --string chassis-type")
     if computer_type in ["Notebook", "Laptop", "Convertible", "Portable"]:
-        # AC adapter states: 0, 1, unknown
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
LGTM! Thank you your contribution and for chasing this one, much appreciated!

MSCHELAK:
I have same issu and I have lenovo laptop

MSCHELAK:
It happen becose conservation_mode

AdnanHodzic:
@MSCHELAK can we please keep the discuss for problems that arose from this in #281? Thanks

Issue #281, opened 2021-11-15
Battery charging state is not recognized


Angelman81:
Hi,

since some days, auto-cpufreq acts like it is not able to identify if the Battery is in charging or discharging mode.

I recognized this by slow CPU speed, the system was set to the "powersave" governor and  auto-cpufreq always showed "Battery is: discharging", but the power adapter was plugged in and the notebook power led reacted to the power cord status. Also the battery task bar symbol of the window manager mate switched between "charging" and "discharging". 

```
andre@andre-ThinkPad-T431s:~$ cat /sys/class/power_supply/AC/online     <--power cord plugged in 
1
andre@andre-ThinkPad-T431s:~$ cat /sys/class/power_supply/AC/online     <--power cord removed
0
```

Of course I did the regular packet updates of the system, but I tried to go back from the fresh 5.11.0-40 to 5.11.0-38, installed auto-cpufreq via the Installer  and then again via Snap, did some reboots, nothing changed.

I have helped myself with the config file and set the  governor for [battery] to "schedutil", thats fine for me. 

I can not rule out that something is wrong with my system, but I did no changes affecting the acpi / battery stuff lately as far as I remember.

---

### System information:
andre@andre-ThinkPad-T431s:~$ sudo auto-cpufreq --debug
[sudo] Passwort für andre:          

Using settings defined in /etc/auto-cpufreq.conf file

-------------------------------------------------------------------------------

Linux distro: Linux Mint 20.2 uma
Linux kernel: 5.11.0-38-generic
Processor: Intel(R) Core(TM) i5-3337U CPU @ 1.80GHz
Cores: 4
Architecture: x86_64
Driver: intel_cpufreq

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 1700 MHz
CPU min frequency: 800 MHz

Core	Usage	Temperature	Frequency
CPU0:	  1.0%     43 °C      800 MHz
CPU1:	  1.0%     43 °C     1700 MHz
CPU2:	  1.0%     42 °C     1620 MHz
CPU3:	  1.0%     42 °C      800 MHz

auto-cpufreq version:
fatal: Kein Git-Repository (oder irgendeines der Elternverzeichnisse): .git
CalledProcessError(128, ['git', 'describe', '--always'])

Python: 3.8.10
psutil package: 5.5.1
platform package: 1.0.8
click package: 7.0
distro package 1.4.0

Computer type: Notebook
Battery is: discharging

[...]

PR #285, opened 2021-11-18
Fix instances where charging is not recognized (#281)

AmitGolden:
Hi,
This PR includes fixes for #281.
Reading from both battery and AC adapter statuses to check if charging.
Also if battery state is unknown is now read as charging


AdnanHodzic:
LGTM! Thank you for your contribution and addressing this problem accordingly.

PR #286, opened 2021-11-20
fixed the heating problem on charging

achintya-7:
### Earlier even though setting the governer as powersave, I still was facing the heating issues on charging because somehow it was not shutting off the turbo. 

debug

stats

config

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -168,10 +168,13 @@ def charging():
 
     computer_type = getoutput("dmidecode --string chassis-type")
     if computer_type in ["Notebook", "Laptop", "Convertible", "Portable"]:
+        # AC adapter states: 0, 1, unknown
+        ac_info = getoutput(f"grep . {power_dir}A*/online").splitlines()
         # Battery statuses: Full, Charging, Discharging, Unknown
         battery_status = getoutput(f"grep . {power_dir}B*/status").splitlines()
-        # if there's one battery charging, ac_state is True
-        ac_state = any(["Full" in ac or "Charging" in ac for ac in battery_status])
+        # if there's one battery charging, or if there's one ac-adapter on-line, ac_state is True
+        ac_state = (any([not "Discharging" in ac for ac in battery_status]) or 
+                    any(["1" in ac.split(":")[-1] for ac in ac_info]))
     else:
         has_battery = psutil.sensors_battery() is not None
         if has_battery:
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -183,6 +183,11 @@ def charging():
         else:
             ac_state = True
 
+    heat = get_config()
+    
+    if(heat['heating_problem']['charging_heating_problem'] == 1):
+        ac_state = False
+
     # if both ac-adapter and battery states are unknown default to not charging
     return ac_state
 
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -189,8 +189,8 @@ def charging():
     heat = get_config()
     
     if(heat['heating_problem']['charging_heating_problem'] == 1):
-        ac_state = False
-
+        ac_state = False 
+    
     # if both ac-adapter and battery states are unknown default to not charging
     return ac_state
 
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Hey, thanks for your contribution, but it seems like some things are missing as part of this PR, as soon as I try to run it in `live` mode I get following error:

```
ahodzic@carbon7 ~/code/auto-cpufreq (achintya-7-master)$ sudo auto-cpufreq --live

Note: You can quit live mode by pressing "ctrl+c"

Linux distro: Ubuntu 21.10 impish
Linux kernel: 5.13.0-21-generic
Cores: 8
Architecture: x86_64
Driver: intel_pstate

CPU max frequency: 4600 MHz
CPU min frequency: 400 MHz

CPU0:	  5.9%     61 °C     2000 MHz
CPU1:	  5.1%     61 °C     2000 MHz
CPU2:	  6.8%     59 °C     3784 MHz
CPU3:	  6.9%     60 °C     2000 MHz
CPU4:	  5.1%     61 °C     2000 MHz
CPU5:	  6.0%     61 °C     2000 MHz
CPU6:	  6.9%     59 °C     2000 MHz
CPU7:	  6.1%     60 °C     2000 MHz

  File "/usr/local/bin/auto-cpufreq", line 4, in <module>
    __import__('pkg_resources').run_script('auto-cpufreq==1.0', 'auto-cpufreq')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 651, in run_script
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1455, in run_script
  File "/usr/local/lib/python3.9/dist-packages/auto_cpufreq-1.0-py3.9.egg/EGG-INFO/scripts/auto-cpufreq", line 169, in <module>
  File "/usr/lib/python3/dist-packages/click/core.py", line 829, in __call__
  File "/usr/lib/python3/dist-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1066, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 610, in invoke
  File "/usr/local/lib/python3.9/dist-packages/auto_cpufreq-1.0-py3.9.egg/EGG-INFO/scripts/auto-cpufreq", line 95, in main
  File "/usr/local/lib/python3.9/dist-packages/auto_cpufreq-1.0-py3.9.egg/auto_cpufreq/core.py", line 874, in set_autofreq
  File "/usr/local/lib/python3.9/dist-packages/auto_cpufreq-1.0-py3.9.egg/auto_cpufreq/core.py", line 191, in charging
KeyError: 'heating_problem'
```

achintya-7:
It seems that this is caused because no auto-cpufreq.conf file is formed by default. I should put this in a try catch field to prevent this error. Can you also check the code again including the auto-cpufreq.conf file.

AdnanHodzic:
Changes make sense to me, but I'll need to test them bit more under heavy workload (as I don't have the overheating you have).

But could you make following changes as part of this PR:

1. Make it work when user is not using config file
2. With these changes, auto-cpufreq --live or --stats will state: "No config file being used" even when `/etc/auto-cpufreq.conf` is in place.

achintya-7:
Ok sir, I will be continuing to work and test on this error after the exams.

AnonPol:
@AdnanHodzic I think it's best the tool respects kernel defaults for CPU governor and only modifies the turbo boost to on or off depending on how intensive the activity is on CPU. This "powersave"  governor of CPUFreq can cause massive under-performance while device is on battery and the "performance" governor can cause overheating while device is charging due to both of these governors from my understanding setting the CPU frequency always to minimum and maximum frequency that it can run at. TLP another battery saving tool by default respects the kernel defaults for CPU governor.

AdnanHodzic:
@AnonPol I think it's best to have this kind of discussion under [discussions](https://github.com/AdnanHodzic/auto-cpufreq/discussions) section and not as part of this PR. But short answer is that auto-cpufreq was designed to respect and let the kernel do the heavy lifting. 

However, landscape is filled with various different hardware and specs which means making decisions automatically is not an easy chores. Hence, if something works better for certain users while they define it as part of [custom configuration file](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq) for their needs, I see no problem in that.

AdnanHodzic:
Closed due to inactivity.

PR #294, opened 2021-11-27
Correcting charge status in multi-battery setup

jdhonea:
Currently on a Thinkpad T480 which utilizes a multiple battery setup.
While not on AC power, auto-cpufreq would incorrectly report that the battery state is charging and set the CPU governor to "performance".

Simple fix below should make sure if any battery is currently in discharging state, the AC state is correctly flagged as discharging.


AdnanHodzic:
Good stuff, thank you for your contribution!

PR #302, opened 2021-12-07
Check if lsb_release exists before using it

bobslept:
When installing auto-cpufreq using the auto-cpufreq-installer on my Fedora 35 system it shows me some errors. 

```
./auto-cpufreq-installer: line 10: lsb_release: command not found
./auto-cpufreq-installer: line 11: lsb_release: command not found
./auto-cpufreq-installer: line 12: lsb_release: command not found
```
lsb_release is not installed by default on my system, and that is why it's not found.

This small change will first check if lsb_release is on the system before using it. In the end it behaves the same, but without the command not found messages.


AdnanHodzic:
Great catch and thank you for your contribution!

PR #303, opened 2021-12-08
Verify the use of supported init systems.

bobslept:
When installing on a old Gentoo machine, I found the following issue. Gentoo doesn't use runit, so it would assume it's systemd. Which was not correct. Added an extra check to see if its actually systemd instead of runit, otherwise give unsupported message. 

```
sudo auto-cpufreq --install

--------------------- Deploying auto-cpufreq as a daemon ----------------------

* Turn off bluetooth on boot [skipping] (package providing bluetooth access is not present)

* Deploy auto-cpufreq remove script

------------------ Running auto-cpufreq daemon install script ------------------

* Unsupported init system detected, could not install the daemon


* Please open an issue on https://github.com/AdnanHodzic/auto-cpufreq


----------------- auto-cpufreq daemon installed and running -----------------

To view live stats, run:
auto-cpufreq --stats

To disable and remove auto-cpufreq daemon, run:
sudo auto-cpufreq --remove

```
The message is maybe not really nice, because it still says it's installed and running afterwards. 

I tested this change on the `non-systemd-gnome-power` branch otherwise it didn't run at all. So I think that is good to go!

### DIFF
--- a/scripts/auto-cpufreq-install.sh
+++ b/scripts/auto-cpufreq-install.sh
@@ -55,7 +55,7 @@ if [ "$(ps h -o comm 1)" = "runit" ];then
 		esac
 	fi
 # Install script for systemd
-else
+elif [ "$(ps h -o comm 1)" = "systemd" ];then
     echo -e "\n* Deploy auto-cpufreq systemd unit file"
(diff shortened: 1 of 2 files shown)


AdnanHodzic:
Looks great, thank you for your contribution and looking forward to seeing more of them! :)

PR #304, opened 2021-12-08
Remove package options from cpufreqctl.sh

bobslept:
This removes the package options from the cpufreqctl.sh script. Because I believe they are not being used in this project. This to avoid confusion from people reading the source and then thinking auto-cpufreq will ship with fonts and what not.

### DIFF
--- a/scripts/cpufreqctl.sh
+++ b/scripts/cpufreqctl.sh
@@ -63,12 +63,6 @@ function help () {
   echo "  --min-perf      Current min_perf_pct options"
   echo "  --max-perf      Current max_perf_pct options"
   echo ""
-  echo "Package options"
-  echo "  --install       Install extra components for all users"
-  echo "  --uninstall     Uninstall extra components for all users"
-  echo "  --update-fonts  Update font cache"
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
I originally wanted to keep `cpufreqctl.sh` as close to its original version, but now that auto-cpufreq is also using it's own version it doesn't make any sense to have functionality of this script which is not even used. 

Hence, excellent catch!

bobslept:
Do you want me to investigate some more on the usage of this script and remove some code? I was playing very safe with only the package options.

AdnanHodzic:
> Do you want me to investigate some more on the usage of this script and remove some code? I was playing very safe with only the package options.

Sure, I don't see a reason why not. However, if you're looking for something to pick up, check existing issues, there's **a lot** stuff that's low hanging fruit.

> I mean the core functionality stands I guess, so no direct future usage of functions that aren't used at this point? Is that a correct assumption?

Yes, just make sure to thoroughly figure how it works and what's core functionality that's used in auto-cpufreq context.

bobslept:
Thanks, I will take a look around! :smiley:

Issue #307, opened 2021-12-13
Doesn’t change to perform when plugged in


ShabirK21:
So I have a Lenovo laptop which has conservative mode, when I plug in with it on, auto-cpufreq doesn't switch to performance and stays in powersave scaling

bobslept:
This issue is probably the same as #299
I'm not really sure what is going on, because I don't experience this issue myself. I'm on a lenovo which uses the ideapad_laptop kernel module. But it stays in charging state when I am on 60% when using conservative mode.

ShabirK21:
Then it should be due my system battery status being read as unknown. Thanks anyway

bobslept:
Are you comfortable with git and want to test something out?

ShabirK21:
So i tested out the program, and i'm glad to say it works! when my bat status is unknown, the live status reads it as charging and sets the "performance" & switches to "powersave" when unplugged. Awesome! thanks for the fix

bobslept:
Thank you so much for testing this. Very happy to hear it solved your problem! 
We probably need a little bit more users to test this before we actually merge into master.
Thanks again!

AdnanHodzic:
Thanks to @bobslept the proposed changes have now been merged with `master` branch. Changes will be part of the upcoming 1.8.2 release.  

Until the release is made, changes can be installed by either having latest `master` changes and installing them using [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto-cpufreq-installer) or using `beta` channel of Snap package (`sudo snap install auto-cpufreq --beta`). With that said I'm closing this issue, please re-open it if the issue persists.

Issue #306, opened 2021-12-10
Battery status showing wrong!


Gokul2003g:
### Error output:
```text
Paste here error output
```Battery is not charging but status shows charging.

---

### System information:
```text
Paste here output of the auto-cpufreq --debug

```Linux distro: Ubuntu 21.10 (Impish Indri)
Linux kernel: 5.13.0-22-generic
Processor: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
Cores: 8
Architecture: x86_64
Driver: intel_pstate

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 2800 MHz
CPU min frequency: 400 MHz

Core    Usage   Temperature     Frequency
CPU0:     8.0%     52 °C     2800 MHz
CPU1:    10.0%     52 °C     2467 MHz
CPU2:     9.9%     52 °C     2800 MHz
CPU3:     8.0%     52 °C     2800 MHz
CPU4:     3.2%     52 °C     2800 MHz
CPU5:     7.1%     52 °C     2800 MHz
CPU6:     7.1%     52 °C     2800 MHz
CPU7:     7.9%     52 °C     2800 MHz

auto-cpufreq version:
Snap: 1.8.0

Python: 3.8.10
psutil package: 5.8.0
platform package: 1.0.8
click package: 8.0.3
Traceback (most recent call last):
  File "/snap/auto-cpufreq/89/bin/pip3", line 5, in <module>
    from pip._internal.cli.main import main
  File "/snap/auto-cpufreq/89/lib/python3.8/site-packages/pip/_internal/cli/main.py", line 10, in <module>
    from pip._internal.cli.autocompletion import autocomplete
  File "/snap/auto-cpufreq/89/lib/python3.8/site-packages/pip/_internal/cli/autocompletion.py", line 9, in <module>
    from pip._internal.cli.main_parser import create_main_parser
  File "/snap/auto-cpufreq/89/lib/python3.8/site-packages/pip/_internal/cli/main_parser.py", line 7, in <module>
    from pip._internal.cli import cmdoptions
  File "/snap/auto-cpufreq/89/lib/python3.8/site-packages/pip/_internal/cli/cmdoptions.py", line 24, in <module>
    from pip._internal.exceptions import CommandError
  File "/snap/auto-cpufreq/89/lib/python3.8/site-packages/pip/_internal/exceptions.py", line 10, in <module>
    from pip._vendor.six import iteritems
[...]

Issue #299, opened 2021-12-06
Not working properly with Lenovo's battery protection


DaniAndTheWeb:
Hi, 
I noticed when battery protection is active (the battery charges until 60% ) and the laptop is plugged in but not charging the service detects the battery as unplugged and disables the charging governor.
It would be great if there could be a workaround to use the charging governor and the battery protection at the same time

AdnanHodzic:
Hi,

Yes it would be great, however I don't own one such device so I'm not sure how I would test this. If you're willing to give it a try, please consider [contributing by creating a PR](https://github.com/AdnanHodzic/auto-cpufreq/#code-contribution).

bobslept:
This is probably different per Lenovo device. I'm on a Lenovo laptop which uses the ideapad_laptop kernel module. In conservation mode, it stops charging around 60% but still has the charging status.

What is the model you are using?

DaniAndTheWeb:
I'm using the same module.
I should add that right now the behavior came back to being normal even if I hadn't changed any configuration.
Maybe it has been some package conflicting with the service but right now I'm not being able to replicate it.

tnagorra:
For me,
`tlp-stat -b` shows `BAT0` status to be `Idle` but `cat /sys/class/power_supply/BAT0/status` shows `Unknown`

```
~
[I] ➜ sudo tlp-stat -b
--- TLP 1.4.0 --------------------------------------------

+++ Battery Care
Plugin: thinkpad
Supported features: charge thresholds, recalibration
Driver usage:
* natacpi (thinkpad_acpi) = active (charge thresholds)
* tpacpi-bat (acpi_call)  = active (recalibration)
Parameter value ranges:
* START_CHARGE_THRESH_BAT0/1:  0(off)..96(default)..99
* STOP_CHARGE_THRESH_BAT0/1:   1..100(default)

+++ ThinkPad Battery Status: BAT0 (Main / Internal)
/sys/class/power_supply/BAT0/manufacturer                   = LGC
/sys/class/power_supply/BAT0/model_name                     = 5B10W51826
/sys/class/power_supply/BAT0/cycle_count                    =     20
/sys/class/power_supply/BAT0/energy_full_design             =  50500 [mWh]
/sys/class/power_supply/BAT0/energy_full                    =  51480 [mWh]
/sys/class/power_supply/BAT0/energy_now                     =  44580 [mWh]
[...]

PR #308, opened 2021-12-13
Change the way of checking the charging state

bobslept:
Would always be charging, unless a battery tells us we are discharging, or the AC we found tells us we are connected to a cable.

This would also work on systems where somehow we cannot find an AC but only a battery. Which I have seen in some earlier opened issues.

Works on my system with ACAD as AC and BAT1 as battery

Fixes issue #306  
Fixes issue #307 
Fixes issue #299

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -165,31 +165,52 @@ def charging():
     """
     get charge state: is battery charging or discharging
     """
-    power_dir = "/sys/class/power_supply/"
+    power_supply_path = "/sys/class/power_supply/"
+    power_supplies = os.listdir(Path(power_supply_path))
 
-    computer_type = getoutput("dmidecode --string chassis-type")
-    if computer_type in ["Notebook", "Laptop", "Convertible", "Portable"]:
-        # AC adapter states: 0, 1, unknown
-        ac_info = getoutput(f"grep . {power_dir}A*/online").splitlines()
-        # Battery statuses: Full, Charging, Discharging, Unknown
-        battery_status = getoutput(f"grep . {power_dir}B*/status").splitlines()
-        # if there's one battery charging, or if there's one ac-adapter on-line, ac_state is True
-        ac_state = (all([not "Discharging" in ac for ac in battery_status]) or 
-                    any(["1" in ac.split(":")[-1] for ac in ac_info]))
+    # check if we found power supplies. on a desktop these are not found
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Thanks for looking into this! This change was originally started as part of #168 based on a [lot of user feedback](https://twitter.com/fooctrl/status/1357230198857564160).

Proposed changes looks sound and they also worked on my side (X1 Carbon), I just wish if we had more confirmations that it works on various/different hardware. It would be great if battery detection issues could be solved once and for all, as besides the issues you mentioned, these issues keep popping up.

Let me think about this and in meantime let's wait for confirmations in #307 and #299.

bobslept:
Yes, I totally understand. I would love to see more way users test this before actually merged into master.

bobslept:
Note for this PR after reading alot of ealier issues. As seen in #124

"hid" devices may show up under /sys/class/power_supply/ as discharging battery.
It should be no problem to filter any item that contains "hid" before actually looping all the items. 
Or just skip them while looping.

bobslept:
This is really great. Already all the issues are reported as fixed. I have told that we need probably need a little bit more testing before actually merging this. 

One other thing that makes me very happy. One person mentioned that the PR #285 had fixed an issue for them, and had positive results with this current PR also. Which is great, because codewise they are totally different. So I think we are on the right track!

AdnanHodzic:
> 

This is definitely the step in right direction. Here's an idea, what if we use logic you implemented as part of this PR as "main" logic, and as a "fallback" (at the end of loop in case nothing was matched) use the old mechanism to determine if device is charging or not?

This way, we could have best of the both worlds. This can also be a transitory thing, where once we've determined that new "main" logic catches every case old mechanism can be completely removed.

bobslept:
I really don't know. Just leave it here and see how it works out.

AdnanHodzic:
@bobslept I've decided to merge these changes with `master`, let's just go with it and if there are numerous people complaining about charging state not being detected properly things can be improved as we go. 

Please do be aware I will tap you if I see this top of issues :slightly_smiling_face:

bobslept:
Really interesting to see what happends. Will be on top of new battery issues :innocent:

AdnanHodzic:
@bobslept since you've been involved with auto-cpufreq a lot as of lately, you might be interested in #312 :slightly_smiling_face:

Issue #276, opened 2021-11-12
issues on pop_os


srulikuk:
I have some issues on pop_os

1. 
```
user@pop-os:~$ sudo auto-cpufreq --monitor
...
...
...
  File "/snap/auto-cpufreq/77/lib/python3.8/site-packages/distro.py", line 1064, in _os_release_info
    with open(self.os_release_file) as release_file:
PermissionError: [Errno 13] Permission denied: '/etc/os-release'
```
Turns out in pop_os the /etc/os-release file is symlinked to another file, as a workaround i removed the symlink and copied the file in its place, you might want to add a check to see if the file is symlinked maybe follow to main file or something.
 
2. this is most likely not the right place to ask but you might know so giving it a shot, pop_os does NOT use grub, instead uses systemd-boot (never used it before) would you know how to change from `intel_pstate` to `acpi-cpufreq` ?

3. for some reason "auto-cpufrq" is not listed in the pop_os store, i installed the snap from the command line however tab wouldnt auto complete the name, and typing in the terminal "auto-cpu" wont auto-complete with tab.

Thanks

AdnanHodzic:
> I have some issues on pop_os
>
>user@pop-os:~$ sudo auto-cpufreq --monitor
...
...
...
  File "/snap/auto-cpufreq/77/lib/python3.8/site-packages/distro.py", line 1064, in _os_release_info
    with open(self.os_release_file) as release_file:
PermissionError: [Errno 13] Permission denied: '/etc/os-release'
Turns out in pop_os the /etc/os-release file is symlinked to another file, as a workaround i removed the symlink and copied the file in its place, you might want to add a check to see if the file is symlinked maybe follow to main file or something.


I've seen this error and it's solution before mentioned in few other issues, like: #240 

>this is most likely not the right place to ask but you might know so giving it a shot, pop_os does NOT use grub, instead uses systemd-boot (never used it before) would you know how to change from intel_pstate to acpi-cpufreq ?
>
>for some reason "auto-cpufrq" is not listed in the pop_os store, i installed the snap from the command line however tab wouldnt auto complete the name, and typing in the terminal "auto-cpu" wont auto-complete with tab.

Sorry but both of these seem to be related strictly to your environment, hence I can't help you with that. I would ask someone from Pop OS side.

andrej-zirko:
[...]

PR #311, opened 2021-12-16
Fix for PopOS os_release problem

tyleraharrison:
Fixes issue #276

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -78,7 +78,20 @@ def get_config(config_file=''):
     return get_config.dict
 
 # get distro name
-dist_name = distro.id()
+try:
+    dist_name = distro.id()
+except PermissionError:
+    # Current work-around for distros like Pop!_OS where symlink causes permission issues
+    print("Warning: Cannot get distro name. Attempting to change symlink to hard link for /etc/os-release")
+    yN = input("Continue? [y/N] ")
+    if yN.lower() == "y":
+        # Backup /etc/os-release
+        os.system("sudo mv /etc/os-release /etc/os-release-backup")
+        # Create hard link to /etc/os-release
+        os.system("sudo ln /etc/pop-os/os-release /etc/os-release")
+    else:
+        print("Aborting...")
+        sys.exit(1)
 
 # display running version of auto-cpufreq
 def app_version():
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -82,7 +82,15 @@ try:
     dist_name = distro.id()
 except PermissionError:
     # Current work-around for distros like Pop!_OS where symlink causes permission issues
-    print("Warning: Cannot get distro name. Attempting to change symlink to hard link for /etc/os-release")
+    print("Warning: Cannot get distro name")
+    if os.path.exists("/etc/pop-os/os-release"):
+        print("Pop!_OS detected")
+        print("Pop!_OS uses a symbolic link for the os-release file, this causes issues \
+            and can be fixed by converting to a hard link")
+        print("Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release")
+    else:
+        print("Aborting...")
+        sys.exit(1)
     yN = input("Continue? [y/N] ")
     if yN.lower() == "y":
         # Backup /etc/os-release
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -90,6 +90,7 @@ except PermissionError:
 
         yN = input("Continue? [y/N] ")
         if yN.lower() == "y":
+            print("Creating hard link for /etc/os-release")
             # Backup /etc/os-release
             os.system("sudo mv /etc/os-release /etc/os-release-backup")
             # Create hard link to /etc/os-release
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -95,7 +95,7 @@ except PermissionError:
             # Create hard link to /etc/os-release
             os.system("sudo ln /etc/pop-os/os-release /etc/os-release")
         else:
-            print("Aborting...")
+            print("Operation aborted by user")
             sys.exit(1)
     else:
         print("Aborting...")
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -88,15 +88,13 @@ except PermissionError:
         print("Pop!_OS uses a symbolic link for the os-release file, this causes issues \
             and can be fixed by converting to a hard link")
         print("Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release")
-    else:
-        print("Aborting...")
-        sys.exit(1)
-    yN = input("Continue? [y/N] ")
-    if yN.lower() == "y":
-        # Backup /etc/os-release
-        os.system("sudo mv /etc/os-release /etc/os-release-backup")
-        # Create hard link to /etc/os-release
-        os.system("sudo ln /etc/pop-os/os-release /etc/os-release")
+
+        yN = input("Continue? [y/N] ")
+        if yN.lower() == "y":
+            # Backup /etc/os-release
+            os.system("sudo mv /etc/os-release /etc/os-release-backup")
+            # Create hard link to /etc/os-release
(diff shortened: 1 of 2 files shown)


bobslept:
Nice! Great improvement for the PopOS users. Maybe before offering this option to the user, check if `/etc/pop-os/os-release` exists so that you are for sure being on the PopOS distro which causes this error?

AdnanHodzic:
> Nice! Great improvement for the PopOS users. Maybe before offering this option to the user, check if `/etc/pop-os/os-release` exists so that you are for sure being on the PopOS distro which causes this error?

Absolutely agree, let's run this only if detected distro is pop os.

tyleraharrison:
How does this look?

auto_cpufreq/core.py
```diff
@@ -78,7 +78,28 @@ def get_config(config_file=''):
...
+    dist_name = distro.id()
+except PermissionError:
+    # Current work-around for distros like Pop!_OS where symlink causes permission issues
+    print("Warning: Cannot get distro name")
+    if os.path.exists("/etc/pop-os/os-release"):
+        print("Pop!_OS detected")
+        print("Pop!_OS uses a symbolic link for the os-release file, this causes issues \
+            and can be fixed by converting to a hard link")
+        print("Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release")
+    else:
```
bobslept, line 91: I would remove these

bobslept, line 91: I would drop this else block

auto_cpufreq/core.py
```diff
@@ -78,7 +78,28 @@ def get_config(config_file=''):
...
+        print("Aborting...")
+        sys.exit(1)
+    yN = input("Continue? [y/N] ")
```
bobslept, line 94: And just add this to the os.path.exists if block

bobslept:
Almost there, I would just do it like this: 

```
try:
    dist_name = distro.id()
except PermissionError:
    # Current work-around for distros like Pop!_OS where symlink causes permission issues
    print("Warning: Cannot get distro name")
    if os.path.exists("/etc/pop-os/os-release"):
        print("Pop!_OS detected")
        print("Pop!_OS uses a symbolic link for the os-release file, this causes issues \
            and can be fixed by converting to a hard link")
        print("Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release")

        yN = input("Continue? [y/N] ")
        if yN.lower() == "y":
            # Backup /etc/os-release
            os.system("sudo mv /etc/os-release /etc/os-release-backup")
            # Create hard link to /etc/os-release
            os.system("sudo ln /etc/pop-os/os-release /etc/os-release")
    else:
        print("Aborting...")
        sys.exit(1)
```

I've tried to comment the code as review kind of thing on github, but never did that before. So I think this is a little bit easier for now. But I will try to add my comments via review thingy, just as a test.

auto_cpufreq/core.py
```diff
@@ -78,7 +78,28 @@ def get_config(config_file=''):
...
+try:
+    dist_name = distro.id()
+except PermissionError:
+    # Current work-around for distros like Pop!_OS where symlink causes permission issues
+    print("Warning: Cannot get distro name")
+    if os.path.exists("/etc/pop-os/os-release"):
+        print("Pop!_OS detected")
+        print("Pop!_OS uses a symbolic link for the os-release file, this causes issues \
+            and can be fixed by converting to a hard link")
+        print("Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release")
```
bobslept, line 90: I would just remove the else block below

tyleraharrison:
I too have never been able to do code revisions in pull requests, been meaning to learn. Will revise

tyleraharrison:
With this modification users are still able to continue even though the OS name was never read if they press Y. Is that intended behavior?

tyleraharrison:
dist_name will not be initialized and cause issues at `elif dist_name in ["arch", "manjaro", "garuda"]:`

bobslept:
No now your `Continue? [y/N]` is part of the `if os.path.exists("/etc/pop-os/os-release"):` code.
So if the file exists, you will ask them to continue, if that file did not exists, it will go to the `print("Aborting...")`

tyleraharrison:
I was just looking at that, didn't notice it got indented. Making the change now :+1:

bobslept:
Great I think we are there! Now we just wait for @AdnanHodzic to review it one last time and then probably merge it when he is happy! 

I will already say thank you for your contribution to auto-cpufreq!  :+1:

tyleraharrison:
Glad I could help, this tool has served me well on my new Framework laptop so I thought I'd give back.

auto_cpufreq/core.py
```diff
@@ -78,7 +78,26 @@ def get_config(config_file=''):
...
+
+        yN = input("Continue? [y/N] ")
+        if yN.lower() == "y":
+            # Backup /etc/os-release
+            os.system("sudo mv /etc/os-release /etc/os-release-backup")
+            # Create hard link to /etc/os-release
+            os.system("sudo ln /etc/pop-os/os-release /etc/os-release")
```
bobslept, line 97:
```suggestion
        if yN.lower() == "y":
            # Backup /etc/os-release
            os.system("sudo mv /etc/os-release /etc/os-release-backup")
            # Create hard link to /etc/os-release
            os.system("sudo ln /etc/pop-os/os-release /etc/os-release")
        else:
            # choose N so abort
            print("Operation aborted")
            sys.exit(1)
```

bobslept (commented):
Well I've just spotted that we do nothing, when we chose N option. So maybe it's a good idea to actually handle it.

tyleraharrison:
Did not mean to hit resolve, let me open the code real quick

bobslept:
I don't really know if I did that right. But the thing is. If the user choose N. we need also nicely print abort and exit.

tyleraharrison:
Is there a clean way to do that without just adding an identical abort block?

tyleraharrison:
```
    yN = input("Continue? [y/N] ")
        if yN.lower() == "y":
            # Backup /etc/os-release
            os.system("sudo mv /etc/os-release /etc/os-release-backup")
            # Create hard link to /etc/os-release
            os.system("sudo ln /etc/pop-os/os-release /etc/os-release")
        else:
            print("Aborting...")
            sys.exit(1)
    else:
        print("Aborting...")
        sys.exit(1)
```
        
Seems a little dirty but it gets the job done

tyleraharrison:
image

bobslept:
Yeah I see. Not really know a nicer way of doing this at the moment. That's why I added another print message. Something like "Operation aborted by user." Would make more sense on that print maybe.

tyleraharrison:
That should do it

auto_cpufreq/core.py
```diff
@@ -78,7 +78,28 @@ def get_config(config_file=''):
...
+except PermissionError:
+    # Current work-around for distros like Pop!_OS where symlink causes permission issues
+    print("Warning: Cannot get distro name")
+    if os.path.exists("/etc/pop-os/os-release"):
+        print("Pop!_OS detected")
+        print("Pop!_OS uses a symbolic link for the os-release file, this causes issues and can be fixed by converting to a hard link")
+        print("Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release")
+
+        yN = input("Continue? [y/N] ")
+        if yN.lower() == "y":
```
bobslept, line 92:
```suggestion
        if yN.lower() == "y":
            print("Creating hard link for /etc/os-release")
```

bobslept (commented):
What do you think as last change. So actually printing it's doing something after choosing "y"

auto_cpufreq/core.py
tyleraharrison: After pressing "commit changes" it does nothing 🤦‍♂️

bobslept:
Wow that's was fast! Well I think it's great! Thank you sir :+1:

tyleraharrison:
Yeah now I understand how the "commit changes" works, just didn't see the indicator for a bit on my side

AdnanHodzic:
Looks great, thank you for your contribution @tyleraharrison and thank you for your assistance reviewing this PR @bobslept!

PR #314, opened 2021-12-18
Support for openrc

AdwaitAdk:
Added support for openrc.


bobslept:
Very nice! Tested on an old Gentoo box which uses openrc as init system and it works for me.

AdnanHodzic:
@AdwaitAdk awesome! Thank you for your contribution! 

@bobslept thank you for your assistance!

PR #317, opened 2021-12-20
Rename various init scripts

bobslept:
We currently have a couple of init files in the `scripts` directory.  To be more clear what the file is, I renamed them. Also changed the corresponding install script lines.


AdnanHodzic:
Good one, thanks!

AdnanHodzic:

PR #319, opened 2021-12-20
Fix removal of shared_dir on --remove

bobslept:
When using `sudo ./auto-cpufreq-installer --remove` the script doesn't actually remove our 'shared_dir' at `/usr/local/share/auto-cpufreq/`

This is now fixed.


AdnanHodzic:
Another great catch, thanks!

AdnanHodzic:

PR #325, opened 2021-12-21
Reformatted codebase

varaki:
All python scripts are now reformatted with 'black' using line length
of 100.

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -17,7 +17,7 @@ from pathlib import Path
 from shutil import which
 from subprocess import getoutput, call, run, check_output, DEVNULL
 
-sys.path.append('../')
+sys.path.append("../")
 from auto_cpufreq.power_helper import *
 
 warnings.filterwarnings("ignore")
@@ -65,8 +65,9 @@ def file_stats():
     auto_cpufreq_stats_file = open(auto_cpufreq_stats_path, "w")
     sys.stdout = auto_cpufreq_stats_file
 
-def get_config(config_file=''):
-    if not hasattr(get_config, 'dict'):
+
+def get_config(config_file=""):
+    if not hasattr(get_config, "dict"):
         get_config.dict = dict()
 
         config = configparser.ConfigParser()
@@ -77,6 +78,7 @@ def get_config(config_file=''):
 
     return get_config.dict
 
+
 # get distro name
 try:
     dist_name = distro.id()
@@ -85,8 +87,12 @@ except PermissionError:
     print("Warning: Cannot get distro name")
     if os.path.exists("/etc/pop-os/os-release"):
         print("Pop!_OS detected")
-        print("Pop!_OS uses a symbolic link for the os-release file, this causes issues and can be fixed by converting to a hard link")
-        print("Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release")
+        print(
+            "Pop!_OS uses a symbolic link for the os-release file, this causes issues and can be fixed by converting to a hard link"
+        )
+        print(
+            "Attempting to change symlink to hard link for /etc/os-release -> /etc/pop-os/os-release"
+        )
 
         yN = input("Continue? [y/N] ")
         if yN.lower() == "y":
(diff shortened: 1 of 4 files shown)


AdnanHodzic:
Thanks for your contribution! 

I'm curious how and why did you settle on 100 instead of i.e: 80?

varaki:
> Thanks for your contribution!
> 
> I'm curious how and why did you settle on 100 instead of i.e: 80?

80 is just too low, 100 is a better option I think, could be 160 as well, then we can say it's twice as much of what PEP-8 recommends.

AdnanHodzic:
Makes sense and it looks good. 

Was thinking if we're doing this with code as part of this PR I would also change the output to be 100 instead of 80 characters long, i.e: 

```
------------------ Running auto-cpufreq daemon install script ------------------
```

all of these are <= 80 but then most of the space would be empty and it won't look good if terminal is scaled below 100 characters, so let's keep as it is unless you have a better idea.

varaki:
> Makes sense and it looks good.
> 
> Was thinking if we're doing this with code as part of this PR I would also change the output to be 100 instead of 80 characters long, i.e:
> 
> ```
> ------------------------------ Current CPU stats ------------------------------
> ------------------ Running auto-cpufreq daemon install script ------------------
> ```
> 
> all of these are <= 80 but then most of the space would be empty and it won't look good if terminal is scaled below 100 characters, so let's keep as it is unless you have a better idea.

I think we can use a function, that dynamically prints out messages in the above format considering the terminal size.
I've quickly created an example, try it out, it works fine.
You can control the inner padding as well as the decor character(s) to use:

```
#!/usr/bin/env python3

import os

def dynamic_print(message, decor="-", padding=1):
    """
    Dinamically print messages according to available terminal width
    If message is equals to or longer than the terminal width, print the message as is

    Parameters:
        message (str): The message to print
        decor (str):   Decorator character(s)
        padding (int): Padding size (how many whitespaces between decor and message)

    Returns:
        str: Formatted message
    """
    term_width = os.get_terminal_size().columns
    message_size = len(message)

    if message_size >= term_width:
        print(message)
        return

    # calculate decor size per side
    decor_size = int((term_width - (message_size + (padding * 2))) / 2)
    decor = f"{int(decor_size / len(decor)) * decor}"
    pad = " " * padding

    # print the formatted message
    print(f"{decor}{pad}{message}{pad}{decor}")

if __name__ == "__main__":
    dynamic_print("Current CPU stats")
    dynamic_print("Running auto-cpufreq daemon install script")
```

I'll check if there's room for improvement for this, update the scripts to use this and I'll submit this as a separate PR.

AdnanHodzic:
> I think we can use a function, that dynamically prints out messages in the above format considering the terminal size.

This is exactly what I was thinking of and what would be ideal solution but didn't express it as an idea :slightly_smiling_face: 

However, code sample from above didn't dynamically scale anything, and `-` were printed out in multiple line.

varaki:
> However, code sample from above didn't dynamically scale anything, and `-` were printed out in multiple lines.

What do you mean? I don't get it.

I tried it on my work laptop (gnome + gnome terminal),  when I posted the example here and it worked.
Now I'm on my home laptop (swaywm + alacritty), also works fine.
Here's a demo of the command run with a oneliner, for more info:
https://user-images.githubusercontent.com/23281360/147129736-511a3f2a-6002-41bc-b071-31f7ea338a2d.mp4

AdnanHodzic:
> Already printed out lines can't be scaled, so if you run the command in a larger sized terminal, then resize it to a smaller size, then the already printed out lines will be wrapped of course.

This the perfect summary right here. Since auto-cpufreq output is re-outputted every 5 seconds I think it would be great if output could be scaled like this. If you have time to implement this as part of auto-cpufreq and create a PR I'll gladly review it!

varaki:
> > Already printed out lines can't be scaled, so if you run the command in a larger sized terminal, then resize it to a smaller size, then the already printed out lines will be wrapped of course.
> 
> This the perfect summary right here. Since auto-cpufreq output is re-outputted every 5 seconds I think it would be great if output could be scaled like this. If you have time to implement this as part of auto-cpufreq and create a PR I'll gladly review it!

Sure, I'm planning to do that. 🙂

AdnanHodzic:
Awesome, looking forward to that PR then :slightly_smiling_face:

AdnanHodzic:

PR #322, opened 2021-12-20
Fix #321, skip hid devices with battery

bobslept:
Fix for issue #321 

HID device with battery show up under `/sys/class/power_supply`

`hidpp_battery_0`
`hidpp_battery_1`

skip devices contain "hid", tested by user and fixes the issue.

I'm not sure if "hid" is too broad match. that's why I will add those names to the list if future errors come up with this, I can refine if needed.

PR #327, opened 2021-12-21
Ignore hidpp_battery

tgasslander:
When looping through the power supply class, ignore `hidpp_battery`.
A connected wireless mouse will report itself as a type = "battery"
which may trick auto_cpufreq into thinking that the system is running on
batteries.

The issue above was discovered on a ThinkPad P15 running Ubuntu 20.04.


bobslept:
see #322

PR #330, opened 2021-12-22
Added missing package Debian based venv

bobslept:
When install from git using the `auto-cpufreq-installer` I noticed the follow error on a Debian based distro:
```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.
 
    apt install python3.8-venv
```

Added that package to the debian based installer packages.


AdnanHodzic:
Excellent catch! `python3-venv` is the dummy package that should be installed, but I see you set that so :+1:

AdnanHodzic:

PR #324, opened 2021-12-21
Add min/max allowed frequencies option (#309)

varaki:
Setting the minimum/maximum allowed CPU frequencies is now possible
via 'scaling_min_freq' and 'scaling_max_freq' options in config
for both modes (charger and battery).
Values should be given in kHZ.

Example:
scaling_min_freq = 1800000
scaling_max_freq = 2000000

I also did autoformatting via 'black', using --line-length 100.
Code is now a bit tidier than before.

PR #338, opened 2021-12-28
Ability to reset scaling frequencies to default

varaki:
Sets frequencies:

- if option is used in auto-cpufreq.conf: use configured value
- if option is disabled/no conf file used: set default frequencies

Frequency setting is performed only once on power supply change

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -457,8 +457,23 @@ def display_load():
 
 # set minimum and maximum CPU frequencies
 def set_frequencies():
-    conf = get_config()
-    section = "charger" if charging() else "battery"
+    """
+    Sets frequencies:
+     - if option is used in auto-cpufreq.conf: use configured value
+     - if option is disabled/no conf file used: set default frequencies
+    Frequency setting is performed only once on power supply change
+    """
+    power_supply = "charger" if charging() else "battery"
+
+    # don't do anything if the power supply hasn't changed
+    if (
+        hasattr(set_frequencies, "prev_power_supply")
+        and power_supply == set_frequencies.prev_power_supply
+    ):
+        return
+    else:
+        set_frequencies.prev_power_supply = power_supply
+
     frequency = {
         "scaling_max_freq": {
             "cmdargs": "--frequency-max",
@@ -474,9 +489,12 @@ def set_frequencies():
     if not hasattr(set_frequencies, "min_limit"):
         set_frequencies.min_limit = int(getoutput(f"cpufreqctl.auto-cpufreq --frequency-min-limit"))
 
+    conf = get_config()
+
     for freq_type in frequency.keys():
         value = None
-        if not conf.has_option(section, freq_type):
+        if not conf.has_option(power_supply, freq_type):
+            # fetch and use default frequencies
             if freq_type == "scaling_max_freq":
                 curr_freq = int(getoutput(f"cpufreqctl.auto-cpufreq --frequency-max"))
                 value = set_frequencies.max_limit
@@ -487,7 +505,9 @@ def set_frequencies():
                 continue
 
         try:
-            frequency[freq_type]["value"] = value if value else int(conf[section][freq_type].strip())
+            frequency[freq_type]["value"] = (
+                value if value else int(conf[power_supply][freq_type].strip())
+            )
         except ValueError:
             print(f"Invalid value for '{freq_type}': {frequency[freq_type]['value']}")
             exit(1)
@@ -506,9 +526,6 @@ def set_frequencies():
         # set the frequency
         print(message)
         run(f"cpufreqctl.auto-cpufreq {args}", shell=True)
-
-    # set function attribute to avoid re-setting freqs unnecessarily
-    set_frequencies.done = True
 
 
 # set powersave and enable turbo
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Thanks for looking into this so quickly. Changes look good to me, 

Regarding your remarks in #324 

> How do we know whether the scaling frequencies were set by auto-cpufreq before or not?

This is the tricky part, we don't know about it, only way would be to keep saving them in another file (or as env variable) and then use that file/env var as a reference, but this is very tricky and complicated. Although, this seems to be fixed with your latest changes on `varaki-scaling_reset`, correct me if I'm wrong?

> Configuration file gets parsed only on startup and it's not being watched for changes during auto-cpufreq is running.
> Should we always check whether the scaling min/max freqs differ from the defaults?

I would do this, as currently this is very confusing. User changes something, expects to see changes but nothing happens. If you're testing these changes with `auto-cpufreq --live`, then it works perfectly, as you have to close (auto-cpufreq) and start over to see the changes. 

If `auto-cpufreq --install` is used, then only option is to either `auto-cpufreq --remove` and then run `auto-cpufreq --install` again and everything will be fine. 

The way I see it, we have 2 options:

1. if it's detected changed were made to whole `/etc/auto-cpufreq.conf` file or only `scaling_min/max_freq` values, to inform user to: reinstall auto-cpufreq `daemon` or simply restart the computer to apply the changes? (simple, but ugly)
2. Try to do this automatically without user noticing anything and applying the changes (complicated)

>I'll figure out something that's resource friendly and doesn't do any checks/settings more than it's necessary, but looking at >auto-cpufreq as a whole, I think re-organizing it into a class based structure would help a lot in general for long-term.

This is something I have in my private notes/ToDo but never got to it.

varaki:
> Although, this seems to be fixed with your latest changes on varaki-scaling_reset, correct me if I'm wrong?

Yes, the min/max limit values (from cpuinfo_min/max_freq files) are fetched and stored in the corresponding function attributes, when the function gets called first time.
These values will be used as default if not overridden via auto-cpufreq.conf.

For example:

Machine is on battery power, only the "scaling _**max**_freq" is defined in the "**battery**" section, and no frequency limits are defined for "charging" section:

1. the value defined in the config as "scaling_ **max**_freq" is set on the system as "scaling _**max**_freq"
2. the default (fetched from cpuinfo _**min**_freq) value will be set for "scaling _**min**_freq"
3. nothing will be performed again until script is restarted or power supply changes
4. user plugs in the charger
5. charger power supply is detected by the function and since no scaling frequency limits are defined in "charger" section, the value fetched from "cpuinfo _**min**_freq" will be set as "scaling _**min**_freq" and the value fetched from "cpuinfo _**max**_freq" will be set as "scaling _**max**_freq"

AdnanHodzic:
> > Although, this seems to be fixed with your latest changes on varaki-scaling_reset, correct me if I'm wrong?
> 
> Yes, the min/max limit values (from cpuinfo_min/max_freq files) are fetched and stored in the corresponding function attributes, when the function gets called first time. These values will be used as default if not overridden via auto-cpufreq.conf.
> 
> For example:
> 
> Machine is on battery power, only the "scaling _**max**_freq" is defined in the "**battery**" section, and no frequency options are defined for "charger" section:
> 
> 1. the value defined in the config as "scaling_ **max**_freq" will be set on the system as "scaling _**max**_freq"
> 2. the default (fetched from cpuinfo _**min**_freq) value will be set for "scaling _**min**_freq"
> 3. nothing will be performed again until script is restarted or power supply changes
> 4. user plugs in the charger
> 5. charger power supply is detected by the function and since no scaling frequency options are defined in "charger" section, the value fetched from "cpuinfo _**max**_freq" will be set as "scaling _**max**_freq"  and the value fetched from "cpuinfo _**min**_freq" will be set as "scaling _**min**_freq"

I missed this during my initial review, this part works perfectly as you explained. 

But it still brings us to this point, as I think average Joe will have no idea why his newly defined values in config file are not picked up. Hence I think it would be perfect if auto-cpufreq could be restarted in case changes to `auto-cpufreq.conf` are detected. If implementing this type of logic with restart is too complicated. Then at least print out a message to user that app/or computer needs to be restarted in order for changes to be applied (but this pretty ugly solution, as in case they installed it, they'll have no idea how to restart an app).

> > Configuration file gets parsed only on startup and it's not being watched for changes during auto-cpufreq is running.
> > Should we always check whether the scaling min/max freqs differ from the defaults?
> 
> I would do this, as currently this is very confusing. User changes something, expects to see changes but nothing happens. If you're testing these changes with `auto-cpufreq --live`, then it works perfectly, as you have to close (auto-cpufreq) and start over to see the changes.
> 
> If `auto-cpufreq --install` is used, then only option is to either `auto-cpufreq --remove` and then run `auto-cpufreq --install` again and everything will be fine.
> 
> The way I see it, we have 2 options:
> 
> 1. if it's detected changed were made to whole `/etc/auto-cpufreq.conf` file or only `scaling_min/max_freq` values, to inform user to: reinstall auto-cpufreq `daemon` or simply restart the computer to apply the changes? (simple, but ugly)
> 2. Try to do this automatically when changes to config file are detected (restart auto-cpufreq?) without user noticing anything and applying the changes (complicated)

varaki:
All previously existing param changes required script/service restart anyway, but nevertheless, this can be done in a separate PR, because it's a new feature and?a rather big change.

AdnanHodzic:
> All previously existing param changes required script/service restart to make effect anyway, but nevertheless, this can be done in a separate PR, because it's a new feature and a rather big change.

Makes sense, let's do that then. Thanks!

AdnanHodzic:

PR #326, opened 2021-12-21
Implement python virtual environment

bobslept:
Most of this work is done by @aroundthfur in PR #263 

Based on his PR I've added a few changes.

Seems to work on Fedora 35 with systemd
Seems to work on Gentoo with openrc
Seems to work on Void with runit

Please see if it works for you! :tada: 

**Not tested with Snap**

PR #347, opened 2022-01-06
Fixed the --version command when installed from source

### LINKED ISSUE #344: Current version not showing when installed via source code
### Error output:
```
auto-cpufreq version:
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
CalledProcessError(128, ['git', 'describe', '--always'])
```
#### Additional info
The current version is not showing up. I pulled the code via git into my `/home` and installed it from there. Is there an option to tell the current installed version where to find the git repo and get the info regarding the version. Moreover, I wonder what would be the proper way to update the software and if it would be better to put the git repo into `/opt/auto-cpufreq`.

---

### System Information
```
Linux distro: Debian GNU/Linux 11 bullseye
Linux kernel: 5.10.0-9-amd64
Processor: Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
Cores: 12
Architecture: x86_64
Driver: intel_pstate

CPU max frequency: 2900 MHz
CPU min frequency: 800 MHz

CPU0:	  1.1%     40 °C      900 MHz
CPU1:	  1.0%     37 °C      900 MHz
CPU2:	  2.0%     38 °C      900 MHz
CPU3:	  8.8%     37 °C      900 MHz
CPU4:	  0.0%     38 °C      900 MHz
CPU5:	  2.9%     38 °C      900 MHz
CPU6:	  1.0%     40 °C      900 MHz
CPU7:	  0.0%     37 °C      900 MHz
CPU8:	  1.0%     38 °C      900 MHz
CPU9:	  1.0%     37 °C      900 MHz
CPU10:	  3.0%     38 °C      900 MHz
CPU11:	  0.0%     38 °C      900 MHz

auto-cpufreq version:
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
CalledProcessError(128, ['git', 'describe', '--always'])

Python: 3.9.2
psutil package: 5.9.0
platform package: 1.0.8
click package: 8.0.3
distro package 1.6.0

Computer type: Notebook
Battery is: discharging

auto-cpufreq system resource consumption:
cpu usage: 0.0 %
memory use: 0.07 %

Total CPU usage: 3.0 %
Total system load: 0.92
Average temp. of all cores: 38.0 °C 

Currently using: powersave governor
Currently turbo boost is: off
```
---

ariasmn:
Closes #344

The command only worked when running it inside the git repository.

Now, it works from everywhere, and the version is constructed using the latest git tag and the SHA (8 first symbols) of the latest commit, which may help with further debugging in future issues (since when installing from source you don't install a fixed version).

To do so, I have included [this setuptools plugin](https://github.com/dolfinus/setuptools-git-versioning). The template for the versioning [can be changed](https://setuptools-git-versioning.readthedocs.io/en/latest/schemas/index.html) by modifying the options in `setup.py`, but IMO this is a good approach, since, as I previously said, the version is constructed with the latest Git tag as well as the SHA of the latest commit, which may come in handy for debugging future specific issues.

Waiting for your feedback!

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -12,6 +12,7 @@ import time
 import click
 import warnings
 import configparser
+import pkg_resources
 from math import isclose
 from pathlib import Path
 from shutil import which
@@ -111,7 +112,7 @@ except PermissionError:
 # display running version of auto-cpufreq
 def app_version():
 
-    print("auto-cpufreq version:")
+    print("auto-cpufreq version: ", end="")
 
     # snap package
     if os.getenv("PKG_MARKER") == "SNAP":
@@ -120,19 +121,13 @@ def app_version():
     elif dist_name in ["arch", "manjaro", "garuda"]:
         aur_pkg_check = call("pacman -Qs auto-cpufreq > /dev/null", shell=True)
         if aur_pkg_check == 1:
-            print(
-                "Git commit:",
-                check_output(["git", "describe", "--always"]).strip().decode(),
-            )
+            print(pkg_resources.require("auto-cpufreq")[0].version)
         else:
             print(getoutput("pacman -Qi auto-cpufreq | grep Version"))
     else:
         # source code (auto-cpufreq-installer)
         try:
-            print(
-                "Git commit:",
-                check_output(["git", "describe", "--always"]).strip().decode(),
-            )
+            print(pkg_resources.require("auto-cpufreq")[0].version)
         except Exception as e:
             print(repr(e))
             pass
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -121,17 +121,27 @@ def app_version():
     elif dist_name in ["arch", "manjaro", "garuda"]:
         aur_pkg_check = call("pacman -Qs auto-cpufreq > /dev/null", shell=True)
         if aur_pkg_check == 1:
-            print(pkg_resources.require("auto-cpufreq")[0].version)
+            print(get_formatted_version())
         else:
             print(getoutput("pacman -Qi auto-cpufreq | grep Version"))
     else:
         # source code (auto-cpufreq-installer)
         try:
-            print(pkg_resources.require("auto-cpufreq")[0].version)
+            print(get_formatted_version())
         except Exception as e:
             print(repr(e))
             pass
 
+# return formatted version for a better readability
+def get_formatted_version():
+    literal_version = pkg_resources.require("auto-cpufreq")[0].version
+    splitted_version = literal_version.split("+")
+    formatted_version = splitted_version[0]
+    
+    if len(splitted_version) > 1:
+        formatted_version += " (git: " + splitted_version[1] + ")"
+
+    return formatted_version
 
 def app_res_use():
     p = psutil.Process()
--- a/setup.py
+++ b/setup.py
@@ -12,12 +12,16 @@ def read(name):
     with open(os.path.join(this, name)) as f:
         return f.read()
 
+# Used for the tar.gz/snap releases
+VERSION = "1.9.0"
 
 setup(
     name="auto-cpufreq",
-    version_config={
-        "template": "{tag}.{sha}",
-        "dev_template": "{tag}.{sha}"
+    setuptools_git_versioning={
+        "starting_version": VERSION,
+        "template": "{tag}+{sha}",
+        "dev_template": "{tag}+{sha}",
+        "dirty_template": "{tag}+{sha}.post{ccount}.dirty"
     },
     setup_requires=["setuptools-git-versioning"],
     description="Automatic CPU speed & power optimizer for Linux",
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,11 @@ def read(name):
 
 setup(
     name="auto-cpufreq",
-    version="1.0",
+    version_config={
+        "template": "{tag}.{sha}",
+        "dev_template": "{tag}.{sha}"
+    },
+    setup_requires=["setuptools-git-versioning"],
     description="Automatic CPU speed & power optimizer for Linux",
     long_description=readme,
     author="Adnan Hodzic",
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,5 @@
 name: auto-cpufreq
 base: core20
-version: '1.9.0'
 summary: Automatic CPU speed & power optimizer for Linux
 description: |
   Automatic CPU speed & power optimizer for Linux based on active
@@ -11,6 +10,7 @@ description: |
 license: LGPL-3.0
 grade: stable
 confinement: strict
+adopt-info: auto-cpufreq
 
 compression: lzo
 
@@ -27,6 +27,9 @@ parts:
        - coreutils
        - dmidecode
     source: .
+    override-pull: |
+      snapcraftctl pull
+      snapcraftctl set-version `grep ^VERSION $SNAPCRAFT_PART_SRC/setup.py | sed 's/.*"\(.*\)"/\1/'`
 
   deploy-scripts:
     plugin: dump


AdnanHodzic:
Changes, look great!  Before I merge this, couple of questions.

1. Would it be possible for output to be: `auto-cpufreq version: 1.9.0 (git: e913d095)`, as I think it would make more sense?

2. For snap package, could an exception be made where it would print out `version:` line from `auto-cpufreq/snap/snapcraft.yaml` file and also include i.e: `(git: e913d095)`?

Snap can be detected by using:

```
if os.getenv("PKG_MARKER") == "SNAP":
# code here ...
```

bobslept:
I really like this, very nice! :+1: 

One question, does this method also work on a release.tar.gz?

ariasmn:
@AdnanHodzic 
> Changes, look great! Before I merge this, couple of questions.
> 
> 1. Would it be possible for output to be: `auto-cpufreq version: 1.9.0 (git: e913d095)`, as I think it would make more sense?
> 2. For snap package, could an exception be made where it would print out  whatever number is defined after `version:` line from `auto-cpufreq/snap/snapcraft.yaml` file and also include i.e: `(git: e913d095)`?
> 
> Snap can be detected by using:
> 
> ```
> if os.getenv("PKG_MARKER") == "SNAP":
> # code here ...
> ```
1. Not a problem, that can be easily changed.
2. That's hard. I'm not very fond of Snap, but I have been trying to build and test using a Snapcraft docker image. However, I see that the current implementation (with just the changes I made) throw some errors that make the snap build fail. Can you confirm?
Also, getting the current git commit has to be done in build time. I don't really know how the snap build process works, but you can set the version to be "git", [inside the snapcraft.yaml](https://snapcraft.io/docs/snapcraft-yaml-reference). It uses `git describe`, so I don't think output can be customized sadly.

@bobslept 
> I really like this, very nice! +1
> 
> Does this method also work on a release.tar.gz?
> 
> Tested on 1.9.0 tar.gz and not seems happy.
> 
> ```
> WARNING: The wheel package is not available.
> WARNING: The wheel package is not available.
> WARNING: The wheel package is not available.
> WARNING: The wheel package is not available.
> WARNING: The wheel package is not available.
> fatal: not a git repository (or any parent up to mount point /)
> Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
> /home/sjaak/Downloads/auto-cpufreq-1.9.0/.eggs/setuptools_git_versioning-1.8.0-py3.10.egg/setuptools_git_versioning.py:147: UserWarning: `version_config` option is deprecated since setuptools-git-versioning v1.8.0 and will be dropped in v2.0.0
> Please rename it to `setuptools_git_versioning`
> ```
> 
> Changing it to `setuptools_git_versioning` in setup.py will remove the deprecated warning. But still has the wheel package warnings.
> 
> Version will return `auto-cpufreq version: 0.0.1` That's something to keep in mind then.

It needs to have a Git repository metadata inside. Since the tar.gz are packaged without the Git repository info, it would not work. A [version_callback](https://setuptools-git-versioning.readthedocs.io/en/latest/options/version_callback.html) could be set to get around this.

Also, after testing the Snap builds, I think that this approach might not be the best (?). I don't know if the Snap build can be fixed, but if not, another approach should be taken...

AdnanHodzic:
@ariasmn 

> 1. Not a problem, that can be easily changed.

Then, please push these changes, I want to merge this PR and make a new release :slightly_smiling_face: 

> 2\. That's hard. I'm not very fond of Snap, but I have been trying to build and test using a Snapcraft docker image. However, I see that the current implementation (with just the changes I made) throw some errors that make the snap build fail. Can you confirm?
>     Also, getting the current git commit has to be done in build time. I don't really know how the snap build process works, but you can set the version to be "git", [inside the snapcraft.yaml](https://snapcraft.io/docs/snapcraft-yaml-reference). It uses `git describe`, so I don't think output can be customized sadly.

Yea, I thought this would be the case, but we don't need this as we already have a way to get Snap package version. Before I wrote my previous comment, I built a Snap package with your changes and it all worked fine. If you want to try it yourself you can [follow instructions I outlined](https://github.com/AdnanHodzic/auto-cpufreq/pull/326#issuecomment-999690600) for @bobslept but you can also leave part of testing Snap package to me.

I don't have a problem with leaving current version format to be part of Snap package. As this version will have the source code information based on which that Snap was built. Just make sure to incorporate version information you implemented with `app_version` function on line 112 of core.py. Hence if it was a Snap package, perfect output would be, i,e:


```
auto-cpufreq version: 1.9.0 (git: e913d095)
Snap version: 1.9.0
```

As you see in `app_version` function, snap package version is outputted by:
```
print(getoutput("echo Snap: $SNAP_VERSION"))
```

bobslept:
> @bobslept btw, I just made minor change how snap version is outputted in `auto-cpufreq --debug` and I've noticed driver field is not interpreted properly, is this the case for you as well?
> 
> ```
> Driver: /bin/sh: 1: cpufreqctl.auto-cpufreq: not found
> ```

Yes, same for me.

ariasmn:
> @ariasmn
> 
> > 1. Not a problem, that can be easily changed.
> 
> Then, please push these changes, I want to merge this PR and make a new release slightly_smiling_face
> 
> > 2. That's hard. I'm not very fond of Snap, but I have been trying to build and test using a Snapcraft docker image. However, I see that the current implementation (with just the changes I made) throw some errors that make the snap build fail. Can you confirm?
> > Also, getting the current git commit has to be done in build time. I don't really know how the snap build process works, but you can set the version to be "git", [inside the snapcraft.yaml](https://snapcraft.io/docs/snapcraft-yaml-reference). It uses `git describe`, so I don't think output can be customized sadly.
> 
> Yea, I thought this would be the case, but we don't need this as we already have a way to get Snap package version. Before I wrote my previous comment, I built a Snap package with your changes and it all worked fine. If you want to try it yourself you can [follow instructions I outlined](https://github.com/AdnanHodzic/auto-cpufreq/pull/326#issuecomment-999690600) for @bobslept but you can also leave part of testing Snap package to me.
> 
> I don't have a problem with leaving current version format to be part of Snap package. As this version will have the source code information based on which that Snap was built. Just make sure to incorporate version information you implemented with `app_version` function on line 112 of core.py. Hence if it was a Snap package, perfect output would be, i,e:
> 
> ```
> auto-cpufreq version: 1.9.0 (git: e913d095)
> Snap version: 1.9.0
> ```
> 
> As you see in `app_version` function, snap package version is outputted by:
> 
> ```
> print(getoutput("echo Snap: $SNAP_VERSION"))
> ```

@AdnanHodzic

So, I updated the PR as you can see. Let me guide you through the changes and why I took this approach.

I updated the way the `--version` outputs when a Git commit is linked, like you wanted to.
Now, the version is set in the `setup.py`. This should solve the problem when using the command from a tar.gz release (FYI @bobslept ).

Now, about Snap. I followed your instructions and could test all the changes (thanks for that!). As you can see, some changes have been made to the `snapcraft.yaml` file. This is just to get the version from the `setup.py` file so you don't have to change it twice in two different files.

On the other hand, getting the Snap package to output the latest Git commit is possible, but I don't think it makes sense.
Something like [this](https://github.com/snapcrafters/obs-studio/blob/master/snap/snapcraft.yaml#L592) can be done easily, so the version is like` {tag}.{sha}`, but the version would have the SHA of the commit when using the `snap list` command, as well as in the [Snap Store](https://snapcraft.io/auto-cpufreq). In my opinion, this makes the release number uglier, and, since you control the Snap releases, you already know which commit is the latest in the current Snap build (correct me if I'm wrong).

With that, give me some feedback and let me know if I didn't explain myself well enough.

AdnanHodzic:
@ariasmn could you please resolve the conflict before I do the final review? Thank you

ariasmn:
@AdnanHodzic Done, let me know if you need something

AdnanHodzic:
@ariasmn LGTM and outstanding work, thank you for your contribution and I hope to see more of your PR in future :slightly_smiling_face:

AdnanHodzic:

PR #350, opened 2022-01-07
Stop power-profiles-daemon in live mode

bobslept:
When using `auto-cpufreq --live` we did only display a warning about the conflict with `power-profiles-daemon`.
It now actually stops it in live mode, and starts it when live mode is exited.

Also added a check to see if `powerprofilesctl` exists.

### DIFF
--- a/auto_cpufreq/power_helper.py
+++ b/auto_cpufreq/power_helper.py
@@ -95,7 +95,7 @@ def gnome_power_detect_install():
             )
             print("Detected running GNOME Power Profiles daemon service!")
             print("This daemon might interfere with auto-cpufreq and has been disabled.\n")
-            print('Disabled daemon is not automatically disabled in "live" and "monitor" mode and')
+            print('This daemon is not automatically disabled in "monitor" mode and')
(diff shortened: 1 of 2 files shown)


AdnanHodzic:
Great stuff, but it's still using `gnome_power_detect` function which prints out:

```

This daemon might interfere with auto-cpufreq and should be disabled.

Steps to perform this action using auto-cpufreq: power_helper script:
cd auto-cpufreq/auto_cpufreq
python3 power_helper.py --gnome_power_disable
```

When now it should use `gnome_power_detect_install` function which message should now be customized to:

```
print("Detected running GNOME Power Profiles daemon service!")
print("This daemon might interfere with auto-cpufreq and has been disabled.\n")
print('Disabled daemon is not automatically disabled in "monitor" mode and')
print("will be enabled after auto-cpufreq is removed or stopped.")    
```

bobslept:
Pushed the changes like suggested.

AdnanHodzic:
Great, as always thank you for your awesome work and contribution!

AdnanHodzic:

PR #351, opened 2022-01-08
Deploy cpufreqctl on --debug and remove via installer

bobslept:
`cpufreqctl.auto-cpufreq` was deployed everywhere except on `auto-cpufreq --debug`. 

It was only removed when it was deployed as part of the daemon install. 

Now it's always removed when using the `auto-cpufreq-install --remove` option.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -170,6 +170,7 @@ function tool_remove {
   wrapper_script="/usr/local/bin/auto-cpufreq"
   unit_file="/etc/systemd/system/auto-cpufreq.service"
   venv_path="/opt/auto-cpufreq"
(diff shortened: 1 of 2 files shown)


bobslept:
I don't know what happend, but it seems that you already merged #347 or something.

Currently on --debug with current master I've got:
```auto-cpufreq version: 1.8.1.000ccf04```

So maybe not merge this branch before fixing that.

AdnanHodzic:
Amazing stuff! Thanks for tracking this down.

> I don't know what happend, but it seems that you already comitted #347 or something. Not sure if that was the plan.

Another excellent catch, I pushed 2 minor changes in the meantime. It seems like I was reviewing changes on #347 branch and then made my changes on top of it and push it all. Either way changes have been reverted, since PR is still open.

AdnanHodzic:

PR #370, opened 2022-03-01
Wrong path for openrc and runit service files

mizzunet:
Following https://github.com/AdnanHodzic/auto-cpufreq/pull/369


mizzunet:
Reason: https://github.com/AdnanHodzic/auto-cpufreq/pull/369#issuecomment-1056611144

Issue #381, opened 2022-03-28
/usr/bin/cpufreqctl.auto-cpufreq is constantly being written to


popsUlfr:
Hi,

while using auto-cpufreq I noticed that it was writing to disk on every loop.

By using strace 
```
sudo strace -f -e trace=file -p <auto-cpufreq-pid> |& grep -e 'open.*WR' -e 'open.*RW'
```
```
openat(AT_FDCWD, "/usr/bin/cpufreqctl.auto-cpufreq", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 5
...
```

I saw it was `/usr/bin/cpufreqctl.auto-cpufreq` getting constantly written to which is weird.

It seems to be this part:
https://github.com/AdnanHodzic/auto-cpufreq/blob/ae1f53c737d9f6462e980bc1614ee6ca1f6b36cd/auto_cpufreq/core.py#L286-L299

https://github.com/AdnanHodzic/auto-cpufreq/blob/ae1f53c737d9f6462e980bc1614ee6ca1f6b36cd/auto_cpufreq/core.py#L299

I'm not entirely sure what should happen here but I don't think `/usr/bin/cpufreqctl.auto-cpufreq` is supposed to be copied in both cases.

---

### System information:

```

-------------------------------------------------------------------------------

Linux distro: Arch Linux rolling n/a
Linux kernel: 5.16.16-zen1-1-zen
Processor: AMD Ryzen 9 5900X 12-Core Processor
Cores: 24
Architecture: x86_64
Driver: acpi-cpufreq

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 3700 MHz
CPU min frequency: 2200 MHz

Core    Usage   Temperature     Frequency
CPU0:     2.0%     44 °C     3700 MHz
CPU1:     1.0%     44 °C     3700 MHz
CPU2:     0.0%     44 °C     3700 MHz
CPU3:     0.0%     44 °C     3700 MHz
CPU4:     0.0%     44 °C     3700 MHz
CPU5:     0.0%     44 °C     3700 MHz
CPU6:     0.0%     44 °C     3700 MHz
CPU7:     0.0%     44 °C     3700 MHz
CPU8:     0.0%     44 °C     3700 MHz
CPU9:     0.0%     44 °C     3700 MHz
CPU10:    1.0%     44 °C     3700 MHz
CPU11:    0.0%     44 °C     3700 MHz
[...]

PR #420, opened 2022-08-12
Don't overwrite /usr/bin/cpufreqctl.auto-cpufreq excessively.

zoitrok:
Looks like the behavior of constantly overwriting it was not intended,
but introduced inadvertently in a91d4ba36e5f42d025b3978ee1178441b52a372f

This should solve #381


AdnanHodzic:
Excellent catch, thank you for your contribution!

Issue #422, opened 2022-08-14
stats outputting wrong number of CPU core count


PROBUBBLE01:
Hello, I've been using auto-cpufreq for a while now and it works fine. But recently when I checked `auto-cpufreq --stats` it incorrectly displays the number of CPU core I have on my computer. I have an AMD ryzen 5 5500U processor which has 6 cores and 12 threads. But the stats says it is 16. It also mentions disabled CPUs from 12 to 15.
It doesn't seem to affect the functionality of the program but the count is incorrect. output of  `--debug` is provided below

### System information:

-------------------------------------------------------------------------------

Linux distro: Arch Linux  
Linux kernel: 5.19.1-arch2-1
Processor: AMD Ryzen 5 5500U with Radeon Graphics
Cores: 16
Architecture: x86_64
Driver: acpi-cpufreq

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 2100 MHz
CPU min frequency: 1400 MHz

Core	Usage	Temperature	Frequency
CPU0:	  1.0%     38 °C     1588 MHz
CPU1:	  0.0%     38 °C     2100 MHz
CPU2:	  0.0%     38 °C     2100 MHz
CPU3:	  0.0%     38 °C     2100 MHz
CPU4:	  0.0%     38 °C     1397 MHz
CPU5:	  1.0%     38 °C     2100 MHz
CPU6:	  0.0%     38 °C     2100 MHz
CPU7:	  0.0%     38 °C     2100 MHz
CPU8:	  0.0%     38 °C     2096 MHz
CPU9:	  0.0%     38 °C     2100 MHz
CPU10:	  0.0%     38 °C     2100 MHz
CPU11:	  0.0%     38 °C     2100 MHz

Disabled CPUs: 12,13,14,15

auto-cpufreq version: Version         : 1.9.5-1

Python: 3.10.5
psutil package: 5.9.1
platform package: 1.0.8
click package: 8.1.3
distro package: 1.7.0

Computer type: Notebook
Battery is: charging

auto-cpufreq system resource consumption:
cpu usage: 0.0 %
memory use: 0.33 %

Total CPU usage: 0.3 %
Total system load: 0.26
Average temp. of all cores: 37.75 °C 

Currently using: performance governor
Currently turbo boost is: off

-------------------------------------------------------------------------------

Additional info: package installed from AUR

AdnanHodzic:
Is the count correct when using different AMD driver? Reference: #398

[...]

PR #423, opened 2022-08-22
Omit the --all flag from calling nproc to get actual number of cores

zoitrok:
--all seems to return the number of threads for processor units on some systems.


zoitrok:
As discussed in issue #422

AdnanHodzic:
Great idea! Thank you for your contribution :)

PR #430, opened 2022-09-07
SunkenHero CPU fan speed display Patch

SunkenHero:
Changes
- displays fan speed of all available fans
- works with all devices, not only with ThinkPad
- removed task from ToDo list

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -27,6 +27,7 @@ from auto_cpufreq.power_helper import *
 warnings.filterwarnings("ignore")
 
 # ToDo:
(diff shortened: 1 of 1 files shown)


SunkenHero (commented):
added fan speed display for all available fans on all devices

SunkenHero:
I know this pull request(not the code) is a big chaos as it is my first one, please forgive

auto_cpufreq/core.py
```diff
@@ -27,6 +27,7 @@
 warnings.filterwarnings("ignore")
 
 # ToDo:
+# - re-enable CPU fan speed display and make more generic and not only for thinkpad
```
SunkenHero, line 30: Removed task from ToDo list

PR #332, opened 2021-12-24
Improve charging detection function

bobslept:
The list of items in power_supply is now sorted. This way the AC is 'always' on top, and checked first. What I noticed is that internal supplies are always use uppercase, and inserted devices all use lowercase. 

So if this theory is correct the AC and internal battery will always be checked first. 

I also noticed the "scope" file under the hid device with battery. Maybe that file could help identify if an battery is from some other device or not. Will see in future bug reports if that is correct and maybe use that.

If this still not works correctly I've implement an ignorelist. Where we can easily add items or parts of the string from devices we want to be ignored. 

Currently in the ignore list is the "hidpp_battery" that came up in every issue so far.

It's just an improvement of what we already have.

Fixes #331

PR #432, opened 2022-09-08
SunkenHero Code optimization & addition of battery percentage

SunkenHero:
A few code optimizations
- Temperatures are more generic
- Now using psutil to get charging status
- Now showing battery percentage

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -204,7 +204,7 @@ def battery_percentage():
     """
     get batery percentage
     """
-    return psutil.sensors_battery().percent
+    return round(psutil.sensors_battery().percent)
 
 
 def get_avail_gov():
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -218,7 +218,7 @@ def charging():
             # If found in ignore list, skip it.
             if ignore_supply:
                 continue
-                
+
             try:
                 with open(Path(power_supply_path + supply + "/type")) as f:
                     supply_type = f.read()[:-1]
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -1072,7 +1072,7 @@ def sysinfo():
     print(f"Processor:{model_name}")
 
     # get core count
-    total_cpu_count = int(getoutput("nproc --all"))
+    total_cpu_count = int(getoutput("nproc"))
     print("Cores:", total_cpu_count)
 
     # get architecture
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -218,6 +218,7 @@ def charging():
             # If found in ignore list, skip it.
             if ignore_supply:
                 continue
+                
             try:
                 with open(Path(power_supply_path + supply + "/type")) as f:
                     supply_type = f.read()[:-1]
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -1119,16 +1119,10 @@ def sysinfo():
                 core = cpu_core[cpu]
                 cpu_temp_index = core_temp_labels.index(f"Core {core}")
                 temp_per_cpu[i] = core_temp["coretemp"][cpu_temp_index].current
-        elif "k10temp" in core_temp:
-            # https://www.kernel.org/doc/Documentation/hwmon/k10temp
-            temp_per_cpu = [core_temp["k10temp"][0].current] * online_cpu_count
-        elif "zenpower" in core_temp:
-            # https://github.com/AdnanHodzic/auto-cpufreq/issues/145#issuecomment-763294009
-            temp_per_cpu = [core_temp["zenpower"][0].current] * online_cpu_count
-        elif "acpitz" in core_temp:
-            temp_per_cpu = [core_temp["acpitz"][0].current] * online_cpu_count
-        elif "thinkpad" in core_temp:
-            temp_per_cpu = [core_temp["thinkpad"][0].current] * online_cpu_count
+        else:
+            temps = list(psutil.sensors_temperatures())
+            for temp in temps:
+                temp_per_cpu = [core_temp[temp][0].current] * online_cpu_count
     except Exception as e:
         print(repr(e))
         pass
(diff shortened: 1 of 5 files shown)


SunkenHero:
I hope the code isn't too bad, i am not so experienced as i am only 15 years old.

auto_cpufreq/core.py
```diff
@@ -42,9 +42,6 @@
 )
 CPUS = os.cpu_count()
 
-# ignore these devices under /sys/class/power_supply/
-POWER_SUPPLY_IGNORELIST = ["hidpp_battery"]
-
```
SunkenHero, line 47: I think its not needed anymore, but i am not sure

auto_cpufreq/core.py
```diff
@@ -200,60 +197,14 @@ def charging():
...
-        return True
-    # we found some power supplies, lets check their state
-    else:
-        for supply in power_supplies:
-            # Check if supply is in ignore list
-            ignore_supply = any(item in supply for item in POWER_SUPPLY_IGNORELIST)
-            # If found in ignore list, skip it.
-            if ignore_supply:
-                continue
+    return True if psutil.sensors_battery().power_plugged else False
```
SunkenHero, line 201: Using psutil to get charging status

auto_cpufreq/core.py
```diff
@@ -993,10 +944,10 @@ def set_autofreq():
 
     # determine which governor should be used
     if charging():
-        print("Battery is: charging\n")
+        print("Battery: " + str(battery_percentage()) + "%\tcharging\n")
         set_performance()
     else:
-        print("Battery is: discharging\n")
+        print("Battery: " + str(battery_percentage()) + "%\tdischarging\n")
         set_powersave()
```
SunkenHero, line 951: updated to show percentage and charging status

auto_cpufreq/core.py
```diff
@@ -1009,12 +960,12 @@ def mon_autofreq():
...
     if charging():
-        print("Battery is: charging\n")
+        print("Battery: " + str(battery_percentage()) + "%\tcharging\n")
         get_current_gov()
         print(f'Suggesting use of "{get_avail_performance()}" governor')
         mon_performance()
     else:
-        print("Battery is: discharging\n")
+        print("Battery: " + str(battery_percentage()) + "%\tdischarging\n")
         get_current_gov()
```
SunkenHero, line 969: updated to show percentage and charging status

auto_cpufreq/core.py
```diff
@@ -1119,16 +1070,9 @@ def sysinfo():
...
-            # https://github.com/AdnanHodzic/auto-cpufreq/issues/145#issuecomment-763294009
-            temp_per_cpu = [core_temp["zenpower"][0].current] * online_cpu_count
-        elif "acpitz" in core_temp:
-            temp_per_cpu = [core_temp["acpitz"][0].current] * online_cpu_count
-        elif "thinkpad" in core_temp:
-            temp_per_cpu = [core_temp["thinkpad"][0].current] * online_cpu_count
+        else:
+            temp = list(psutil.sensors_temperatures())
+            temp_per_cpu = [core_temp[temp[0]][0].current] * online_cpu_count
     except Exception as e:
```
SunkenHero, line 1076: I made it more generic using the same princip like at the fan speed display

SunkenHero:
Completely tested it on my laptop without encountering any errors

auto_cpufreq/core.py
AdnanHodzic: Good idea, I just hope it remains working on all of that different hardware, let's keep an eye on it based on new issues that come in.

auto_cpufreq/core.py
AdnanHodzic: I think we do still need it, check #332 PR for reference.

auto_cpufreq/core.py
```diff
@@ -200,60 +197,14 @@ def charging():
...
-                # could not find type, check the next item
-                continue
 
-    # we cannot determine discharging state, assume we are on powercable
-    return True
+def battery_percentage():
+    """
+    get batery percentage
+    """
+    return psutil.sensors_battery().percent
```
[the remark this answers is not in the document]
AdnanHodzic: Another reference on why it was the way it was done in the past #308

auto_cpufreq/core.py
AdnanHodzic: Wasn't too share about adding the battery percentage but now that I think about it, this could be a good addition, but let's change how it looks a bit from visual point of view.

Instead of having it like (as it is now):

> Battery: 97%	charging

to have it like:

> Battery is: charging (97%)

auto_cpufreq/core.py
AdnanHodzic: As I mentioned in other comment this was all added with #308 as battery state would not be showing properly on certain devices, still thinking if we might need some of it or if `psutil` will be able to pick it all up itself ...

AdnanHodzic:
> I hope the code isn't too bad, i am not so experienced as i am only 15 years old.

This is great, and I encourage you stay on this path and you'll get far in life :)

I reviewed your changes and left few comments on everything, please take a look. 

If you haven't seen it already and if you plan on sticking around for awhile I also encourage you to take a look at: https://github.com/AdnanHodzic/auto-cpufreq#looking-for-developers-and-co-maintainers

SunkenHero:
Ok i will fix the problems and in the future i will check for issues before changing the code

auto_cpufreq/core.py
SunkenHero: I am pretty sure it should work.
I just changed it so it takes the name to get the tame instead of checking for the name to get the temp

SunkenHero:
so i fixed all problems, and i think its ready for the merge

AdnanHodzic:
> so i fixed all problems, and i think its ready for the merge

One last thing, do you have a problem with this on your side now? `CPU fan speed: 0 RPM`

auto_cpufreq/core.py
AdnanHodzic: Let's go for it, but let's also just keep an eye on new issues and if they could be related to changes here.

PR #433, opened 2022-09-09
Switch to the properly maintained AUR package

xd003:
The package at https://aur.archlinux.org/packages/auto-cpufreq-git which has been linked in readme is out of date since long & is actually abondoned . https://aur.archlinux.org/packages/auto-cpufreq is the one being maintained properly at the moment

Also , at the time of writing this the auto-cpufreq-git package doesn't even compile anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,13 @@ In case you encounter any problems with `auto-cpufreq-installer`, please [submit
 
 ### AUR package (Arch/Manjaro Linux)
 
-[AUR package is available](https://aur.archlinux.org/packages/auto-cpufreq-git/) for install. After which `auto-cpufreq` will be available as a binary and you can refer to [auto-cpufreq modes and options](https://github.com/AdnanHodzic/auto-cpufreq#auto-cpufreq-modes-and-options).
+* [Binary Package](https://aur.archlinux.org/packages/auto-cpufreq)
+(For the latest binary release on github)
+* [Git Package](https://aur.archlinux.org/packages/auto-cpufreq-git)
+(For the latest commits/changes) \
+Please note that this git package is currently unmaintained & has issues. Until someone starts maintaining it, use the [manual script installer](https://github.com/AdnanHodzic/auto-cpufreq#auto-cpufreq-installer) if you intend to have the latest changes.
+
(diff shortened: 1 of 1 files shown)


SunkenHero:
At first, a little typo in your comment(First and second link are the same).
The compile error could be my fault because i recently made an pull request i made.
You can try to comment 1149-1151 in /auto-cpufreq/core.py out and then reinstall the package.

AdnanHodzic:
@xd003 could you please make the changes proposed by @SunkenHero so we can merge this change. Thanks

xd003:
Just to update my findings , there's no error if i install auto-cpufreq manully using the installer script - https://github.com/AdnanHodzic/auto-cpufreq#auto-cpufreq-installer . So the issue seems to be somewhere within the auto-cpufreq-git package itself . So i would assume ideally changes need to be done in there & not really in this repo ? Here's the traceback just in case - https://www.klgrth.io/paste/onscc

AdnanHodzic:
> Just to update my findings , there's no error if i install auto-cpufreq manully using the installer script - https://github.com/AdnanHodzic/auto-cpufreq#auto-cpufreq-installer . So the issue seems to be somewhere within the auto-cpufreq-git package itself . So i would assume ideally changes need to be done in there & not really in this repo ? Here's the traceback just in case - https://www.klgrth.io/paste/onscc

Thanks for checking. Yes, in this case it's definitely problem which should be addressed on AUR package maintainers side.

xd003:
I am not really familiar with Arch Package builds , so i have mentioned in readme to use the binary package for releases & the Installer script if someone needs latest changes until the git package has a active maintainer

SunkenHero:
So i really got no idea how AUR packages work and what they really are, but your commit looks good and if it fixes the problem it would be good to get merged

AdnanHodzic:
@xd003 one small change request from my side.

Could you please:

Set these 2 to be on the same line:
> [Binary Package](https://aur.archlinux.org/packages/auto-cpufreq) (For the latest binary release on github)
> [Git Package](https://aur.archlinux.org/packages/auto-cpufreq-git)(For the latest commits/changes )

Remove empty space before ,/. `has issues .` & `maintaining it ,` and add `.` at the end of sentence here `latest changes`

xd003:
Looks good to me now

AdnanHodzic:
> Looks good to me now

Good stuff, thank you for your contribution and you'll be credited for you work as part of upcoming 1.9.6.

Issue #434, opened 2022-09-10
auto-cpufreq keeps failing on Arch Linux


milkycookies100:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? 
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)?
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)?
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)?

### Error output:

```
sudo systemctl status auto-cpufreq
× auto-cpufreq.service - auto-cpufreq - Automatic CPU speed & power optimizer for Linux
     Loaded: loaded (/etc/systemd/system/auto-cpufreq.service; enabled; preset: disabled)
     Active: failed (Result: exit-code) since Sat 2022-09-10 09:34:25 IST; 32min ago
   Duration: 1.300s
   Main PID: 2177 (code=exited, status=1/FAILURE)
        CPU: 302ms

Sep 10 09:34:25 hmm systemd[1]: auto-cpufreq.service: Scheduled restart job, restart counter is at 6.
Sep 10 09:34:25 hmm systemd[1]: Stopped auto-cpufreq - Automatic CPU speed & power optimizer for Linux.
Sep 10 09:34:25 hmm systemd[1]: auto-cpufreq.service: Start request repeated too quickly.
Sep 10 09:34:25 hmm systemd[1]: auto-cpufreq.service: Failed with result 'exit-code'.
Sep 10 09:34:25 hmm systemd[1]: Failed to start auto-cpufreq - Automatic CPU speed & power optimizer for Linux.
```

### System information:
```
lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         48 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  12
  On-line CPU(s) list:   0-11
Vendor ID:               AuthenticAMD
  Model name:            AMD Ryzen 5 4600H with Radeon Graphics
    CPU family:          23
    Model:               96
[...]

PR #435, opened 2022-09-10
Replaced obsolete "egrep" calls with "grep -E"

hingen:
Fixed an issue #434 caused by the recent update to `grep`. Said update results in a `egrep is obsolescent; using grep -E` line to be printed before the actual desired output whenever `egrep` is called.


AdnanHodzic:
LGTM, thank you for your contribution!

Issue #441, opened 2022-10-09
AttributeError: 'NoneType' object has no attribute 'percent'


vinograd-cyber:
~]# auto-cpufreq --monitor

Note: You can quit monitor mode by pressing "ctrl+c"

-------------------------------------------------------------------------------

Linux distro: CentOS Linux 7 Core
Linux kernel: 3.10.0-1160.66.1.el7.x86_64
Processor: Intel(R) Xeon(R) Gold 6230R CPU @ 2.10GHz
Cores: 52
Architecture: x86_64
Driver: intel_pstate

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 4000 MHz
CPU min frequency: 1000 MHz

Core    Usage   Temperature     Frequency
CPU0:    98.0%     72 °C     2728 MHz
CPU1:   100.0%     72 °C     2728 MHz
CPU2:   100.0%     78 °C     2727 MHz
CPU3:   100.0%     76 °C     2728 MHz
CPU4:   100.0%     79 °C     2728 MHz
CPU5:   100.0%     75 °C     2728 MHz
CPU6:   100.0%     74 °C     2720 MHz
CPU7:   100.0%     73 °C     2728 MHz
CPU8:   100.0%     71 °C     2720 MHz
CPU9:   100.0%     80 °C     2720 MHz
CPU10:   99.0%     79 °C     2721 MHz
CPU11:  100.0%     81 °C     2728 MHz
CPU12:  100.0%     79 °C     2728 MHz
CPU13:  100.0%     70 °C     2728 MHz
CPU14:  100.0%     73 °C     2728 MHz
CPU15:  100.0%     74 °C     2728 MHz
CPU16:   99.0%     79 °C     2728 MHz
CPU17:   98.0%     79 °C     2727 MHz
CPU18:  100.0%     82 °C     2728 MHz
CPU19:  100.0%     78 °C     2728 MHz
CPU20:  100.0%     70 °C     2728 MHz
CPU21:  100.0%     74 °C     2729 MHz
CPU22:   99.0%     74 °C     2728 MHz
CPU23:  100.0%     78 °C     2728 MHz
CPU24:  100.0%     77 °C     2728 MHz
CPU25:   98.0%     81 °C     2728 MHz
CPU26:  100.0%     72 °C     2728 MHz
CPU27:  100.0%     72 °C     2728 MHz
CPU28:  100.0%     78 °C     2728 MHz
[...]

PR #442, opened 2022-10-10
Battery Percentage fix

SunkenHero:
Fix for problem in #441 (AttributeError: 'NoneType' object has no attribute 'percent')


AdnanHodzic:

PR #447, opened 2022-10-30
add wheel to dependencies (should speed up install)

johnseekins:
Signed-off-by: John Seekins <john@civiceagle.com>

`wheel` lets `pip` download pre-compiled binary packages, which should make the actual install process a little smoother for people.


SunkenHero:
That's indeed a good idea, if it works we could implement it and look out if it works good.

AdnanHodzic:
Sorry, didn't have time to look at this earlier.

Thank you for your contribution @johnseekins and thank you for taking a look at it @SunkenHero.

PR #455, opened 2022-11-11
Fixed bug where turbo boost will always be on in performance mode

R1D3R175:


AdnanHodzic:
Great thanks! LGTM, but just give me some time to test it to double check if everything works as expected.

@SunkenHero if you have time could you please also verify if it works as expected on your side? Thanks!

SunkenHero:
It worked completely fine for me, it switched to into turbo during a stress-test and back off after.

AdnanHodzic:
Thank you for your contribution @R1D3R175!

PR #459, opened 2022-11-24
Modified Installer

abvee:
Fixed some formatting for the installer and lessened the number of echo calls with here docs, etc.

Now, seperators are done properly, to the width of the terminal and are proper box dashes (as per the Unicode specification):

image

Adjusting the size adjusts the bar:

image

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -225,7 +225,11 @@ function tool_remove {
 }
 
 function ask_operation {
-  echo -e "\n-------------------------- auto-cpufreq installer -----------------------------"
+  MID="$((COLOUMNS / 2))"
+  HEADER=" auto-cpufreq installer "
+  printf "%0.s-" $(seq $((MID-(${#HEADER}/2))))
+  echo -n $HEADER
+  printf "%0.s-" $(seq $((MID-(${#HEADER}/2))))
   echo -e  "\nWelcome to auto-cpufreq tool installer.
   			\nOptions:\n"
   read -p \
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -105,7 +105,7 @@ function manual_install {
 	After which tool is installed, for full list of options run:
 
   	auto-cpufreq
-	_EOF_
+_EOF_
   separator
   cat <<- _EOF_
 
@@ -117,7 +117,7 @@ function manual_install {
   Distribution: $distro
   Release: $release
   Codename: $codename
-  _EOF_
+_EOF_
   separator
 }
 
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -77,32 +77,47 @@ function completed () {
 
 function complete_msg() {
   separator
-  echo -e "\nauto-cpufreq tool successfully installed.\n"
-  echo -e "For list of options, run:\nauto-cpufreq --help"
+  echo -e 	"\nauto-cpufreq tool successfully installed.\n
+  		For list of options, run:\nauto-cpufreq --help"
   separator
 }
 
 function manual_install {
-  echo -e "\nDidn't detect Debian or RedHat based distro.\n"
-  echo -e "To complete installation, you need to:"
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Very nice, I actually wanted to do this myself but never got around it. Two things:

1. Adjusting the size of terminal will result in breaking the lines, i.e:
This is an image

As from what I understood this doesn't happen for you? If I re-run the same command on new re-sized terminal width then line will be adjusted.

2. Why is this only done for `install`, as for example `remove` will not use any of this new functionality. I.e
```
------------------------- auto-cpufreq daemon removed -------------------------
```

But I guess it's related to what I proposed in 4 as this stuff is used outside of installer

3. Let's ident "For list of options, run:" to the left, as now it looks bit off:

```
auto-cpufreq tool successfully installed.

  		For list of options, run:
auto-cpufreq --help
```


5. After we merge this, it would be great if you could make these same changes across auto-cpufreq output where `sep` and `seperator` are being used.

abvee:
> As from what I understood this doesn't happen for you?

Well, lines that are already drawn will not resize, but yes, running the command again in a resized terminal will adjust to the size.

> Why is this only done for install, as for example remove will not use any of this new functionality. 

I uh... didn't see that. I'll make some commits and fix those, they will probably use the same logic as the installer heading :smile: 

> Let's ident "For list of options, run:" to the left, as now it looks bit off:

Will do

> 
    After we merge this, it would be great if you could make these same changes across auto-cpufreq output where sep and seperator are being used.

The only problem with that is that I don't know python very well.... I usually program in c and make small shell scripts. It shouldn't be too hard though. I'll keep you informed if I cannot do it

AdnanHodzic:
If it helps, there's even a snippet that will take care of terminal sizing/padding but was never merged with `master` branch, reference: https://github.com/AdnanHodzic/auto-cpufreq/pull/325#issuecomment-999486895

abvee:
I'll probably make a new branch and PR to merge anything outside the installer.

Let me know if there are any problems !

AdnanHodzic:
> I'll probably make a new branch and PR to merge anything outside the installer.

Definitely a good idea, since there will be lot more changes and it'll be a lot bigger PR.

> I took the liberty to replace Hyphens in the headers to an "equals to" sign (=):
> Let me know if you prefer one over the other, and if there are any other problems dagger

Let's please return these to `---`, I find the output to be "too bulky" with `===`. Ideally these would be lines as they are for `seperator` and could also scale. See if you could make it like that, as that would be excellent. 

> P.S: Thanks for being so nice on my first PRs :)

Most welcome, I'm honored to make your first PR a pleasant experience and too many more! :clinking_glasses:

abvee:
> Ideally these would be lines as they are for seperator and could also scale. See if you could make it like that, as that would be excellent.

> Also let's add an empty line before "Welcome to ..."

Done

> Most welcome, I'm honored to make your first PR a pleasant experience and too many more! beers

Thanks again :D. 
Let's hope I have many more

AdnanHodzic:
I've just noticed a regression introduced as part of #460, so let's resolve this first before this PR is merged.

AdnanHodzic:
@abvee can you please rebase/include changes that are now on `master` branch as part of this branch so I can fully test the changes? As without them I'm still getting:

```
sudo ./auto-cpufreq-installer --remove


────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

Couldn't remove the auto-cpufreq daemon
/usr/local/bin/auto-cpufreq-remove or /usr/bin/auto-cpufreq-remove do not exist.
```

abvee:
> so maybe it's easiest to make the changes as part of this PR?

Okay, will do.

> you'll end up getting the following error:

Okay, I'll keep the error, but remove the other files....

abvee:
the `auto-cpufreq --remove` already prints a command, so I'm not going to reprint it.

I also added a check for the  wrapper script, so it should just work now

AdnanHodzic:
Perfect! Thank you for your contribution and I'm making to be making a v1.9.7 release where you'll get credit for all your work :)

PR #460, opened 2022-11-24
Use usr/local/bin instead of usr/bin for Silverblue and other immutable OSs

gimbles:
Fixes #394

PR #462, opened 2022-11-29
Fix for Regression in #460

abvee:
The uninstall script was checking old locations (`/usr/bin/<file>`) that were changed by @gimbles

When the check to see if `/usr/bin/auto-cpufreq-remove` existed failed (since it's now at `/usr/local/bin/auto-cpufreq-remove`), the removal didn't occur fully and the installer silently goes on until the  success message is wrongly printed.

Just adding  a check for this should make it work now

However, people who downloaded older versions of `auto-cpufreq` will still have those scripts at the old locations (in `/usr/bin`), so I added some variables with an `_old` suffix so people cloning a new repo and running the script can still uninstall their old versions of `auto-cpufreq`.

We could also use `/usr/local/bin` for only Silverblue, which is something I can work on later, but this should fix the script for now.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -167,7 +167,7 @@ function tool_remove {
   srv_install="/usr/local/bin/auto-cpufreq-install"
   srv_install_old="/usr/bin/auto-cpufreq-install"
 
-  srv_remove="/usr/bin/auto-cpufreq-remove"
+  srv_remove="/usr/local/bin/auto-cpufreq-remove"
   srv_remove_old="/usr/bin/auto-cpufreq-remove"
 
   stats_file="/var/run/auto-cpufreq.stats"
@@ -189,7 +189,7 @@ function tool_remove {
   done
 
   # run uninstall in case of installed daemon
-  if [ -f $srv_remove -o -f $srv_remove_old]; then
+  if [ -f $srv_remove -o -f $srv_remove_old ]; then
     eval $tool_proc_rm
   else
 	separator
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -193,7 +193,8 @@ function tool_remove {
     eval $tool_proc_rm
   else
 	separator
-	printf "Couldn't remove auto-cpufreq\n$srv_remove or $srv_remove_old do not exist.\n"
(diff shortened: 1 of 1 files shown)


SunkenHero:
I quickly looked over it and it looks like everything should work.

AdnanHodzic:
> The uninstall script was checking old locations (`/usr/bin/<file>`) that were changed by @gimbles
> 
> When the check to see if `/usr/bin/auto-cpufreq-remove` existed failed (since it's now at `/usr/local/bin/auto-cpufreq-remove`), the removal didn't occur fully and the installer silently goes on until the success message is wrongly printed.
> 
> Just adding a check for this should make it work now
> 
> However, people who downloaded older versions of `auto-cpufreq` will still have those scripts at the old locations (in `/usr/bin`), so I added some variables with an `_old` suffix so people cloning a new repo and running the script can still uninstall their old versions of `auto-cpufreq`.

Great thinking!

> We could also use `/usr/local/bin` for only Silverblue, which is something I can work on later, but this should fix the script for now.

Yea, for now it's fine and I don't mind keeping it under `/usr/local/bin`, as it would be good to have paths as unified as possible.

Regarding your commit, you have a syntax error, please change line 192 to from:

```
  if [ -f $srv_remove -o -f $srv_remove_old]; then
```

to:

```
  if [ -f $srv_remove -o -f $srv_remove_old ]; then
```

I also get following error when I try to remove (older then this PR) auto-cpufreq installed from current `master` branch:

```
sudo ./auto-cpufreq-installer --remove

Couldn't remove the auto-cpufreq daemon
/usr/bin/auto-cpufreq-remove or /usr/bin/auto-cpufreq-remove do not exist.
```

and I'll run into some issue as part of your branch `abvee-460-regression` (even if prior to `--remove` I do `sudo ./auto-cpufreq-installer --remove`) whether auto-cpufreq daemon is installed or not ...

abvee:
> Regarding your commit, you have a syntax error, please change line 192 to from:
My bad, I'll fix it

> I also get following error when I try to remove (older then this PR) auto-cpufreq installed from current master branch:
Like the idiot I am, I forgot to add `local` to the path. I just tested it, it should work now

AdnanHodzic:
Perfect, and thank you for your contribution!

abvee:
> Perfect, and thank you for your contribution!

Thanks :D

Issue #467, opened 2023-01-03
[Feature] Performance Powersave toggler.


niksingh710:
The tools are working fine.
Just have one request it should have a cli way of interaction to switch the governor.

e.g 
currently, my config states that on charging the governor should be `performance` and when not on AC it should be `powersave`

This ramps up my fan while charging.
sometimes I want it to be on powersave even when it is charging to keep my fan silent. and for that I have to change the config and reload auto-cpufreq

I want the feature to be something like 
```bash
auto-cpufreq --force governer powersave    # This should keep the governer to powersave mode no matter what the state is.
auto-cpufreq --force governer performance # This should keep the governer to performance mode no matter what the state is (AC or on Battery)
auto-cpufreq --force reset # resets to the config state.
auto-cpufreq --state # (should return powersave,performance,reseted) So that the user can create a indicator in it's bar config.
```

AdnanHodzic:
This is already possible with [auto-cpufreq config file](https://github.com/AdnanHodzic/auto-cpufreq#2-auto-cpufreq-config-file). 

Otherwise since I'm very limited with available time to do this myself. If you want to give it a try in implementing this functionality, please give it a try and [contribute to the project](https://github.com/AdnanHodzic/auto-cpufreq/#code-contribution) and you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

AdnanHodzic:
Changes made by @shadeyg56 as part of #476 are now merged with `master` branch, hence closing this issue.

Issue #398, opened 2022-05-26
CPU not scaling up in battery mode [Kernel 5.17 with amd-pstate driver]


mldytech:
Hello,

first of all thanks for this very helpful tool.
I just installed kernel 5.17 (including the new amd-pstate driver) on my system and am experiencing weird behaviour when using auto-cpufreq. First of all, the "min" and "max" frequencies seem to adjust dynamically (seen in auto-cpufreq --stats).
I tested the following scenarios while running prime95 (full cpu-load simulation). When charging, the freq goes only up to 2.3 GHz on all cores, and in battery mode I barely reach 1 GHz. I also experienced low frequencies (when on battery) with kernel 5.15 and the acpi driver.

I tried installing auto-cpufreq through the AUR package, as well as from source. 
Am I doing something wrong, or is using the new amd-pstate driver a bad idea in general?

Thank you very much!

<details>
<summary>Output of auto-cpufreq --debug</summary>
### System information:

Base clock = 1.8 GHz 
Boost = up to 4.3 GHz

-------------------------------------------------------------------------------

Linux distro: Manjaro Linux 21.2.6 Qonos
Linux kernel: 5.17.9-1-MANJARO
Processor: AMD Ryzen 7 5700U with Radeon Graphics
Cores: 16
Architecture: x86_64
Driver: amd-pstate

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 1801 MHz
CPU min frequency: 400 MHz

Core	Usage	Temperature	Frequency
CPU0:	  5.8%     51 °C     1801 MHz
CPU1:	  1.0%     51 °C     1776 MHz
CPU2:	  5.0%     51 °C     1801 MHz
CPU3:	  0.0%     51 °C     1801 MHz
CPU4:	  1.0%     51 °C     1801 MHz
CPU5:	  1.0%     51 °C     1801 MHz
CPU6:	  1.0%     51 °C     1801 MHz
CPU7:	  0.0%     51 °C     1801 MHz
CPU8:	  2.0%     51 °C     1801 MHz
CPU9:	  0.0%     51 °C     1801 MHz
CPU10:	  0.0%     51 °C     1801 MHz
CPU11:	  0.0%     51 °C     1801 MHz
CPU12:	  1.0%     51 °C     1801 MHz
CPU13:	  0.0%     51 °C     1801 MHz
CPU14:	  7.4%     51 °C     1801 MHz
[...]

PR #476, opened 2023-01-20
[Feature] Force use of powersave or performance governors

shadeyg56:
I noticed a couple of people had requested this and I wanted the feature myself, so I took the initiative. I threw this together, but I tested it well and it works flawlessly so far.

This adds two new options.

`--force TEXT` allows the user to force a governor into use until the user disables it.
For example, if I do `sudo auto-cpufreq --force=powersave` this will force the use of the powersave governor under any circumstance. If you plug in your laptop it will still use powersave mode, and changing the config file will not do anything. This will even persist after a reboot because it's stored to disk. 

Using `sudo auto-cpufreq --force=reset` will reset the override and return to using the config file or default governors

the `--state` options returns the status of the override. if there's no override it simply returns `default`

Hopefully, this is a welcome change. You're welcome to change any of the help messages or updates I made to the README as I made them on the fly and didn't put a ton of thought into the wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,9 @@ auto-cpufreq should be run with with one of the following options:
 * force TEXT
     - Force use of either the powersave or performance governor. Setting to 'reset' goes back to normal mode
 
+* state
+    - Show if the governor has been overriden with --force. Returns 'default' if there's no override
+
 * config TEXT
     - Use config file at defined path
 
--- a/README.md
+++ b/README.md
@@ -206,13 +206,13 @@ auto-cpufreq should be run with with one of the following options:
 * [stats](https://github.com/AdnanHodzic/auto-cpufreq/#stats)
     - View live stats of CPU optimizations made by daemon
 
-* force TEXT
+* force=TEXT
     - Force use of either the powersave or performance governor. Setting to 'reset' goes back to normal mode
 
 * state
     - Show if the governor has been overriden with --force. Returns 'default' if there's no override
 
-* config TEXT
+* config=TEXT
     - Use config file at defined path
 
 * debug
@@ -242,6 +242,12 @@ No changes are made to the system, and is solely made for demonstration purposes
 `sudo auto-cpufreq --live`
 
 Necessary changes are temporarily made to the system which are lost with system reboot. This mode is made to evaluate what the system would behave with auto-cpufreq permanently running on the system.
+
+### Overriding governor
+
+`sudo auto-cpufreq --force=governor`
+
+The system will be forced to use the governor of choice. The only governor options are "powersave" and "performance". Setting force to "reset" will disable any override. Please note that any set override will persist even after reboot.
 
 ### Install - auto-cpufreq daemon
 
--- a/README.md
+++ b/README.md
@@ -209,7 +209,6 @@ auto-cpufreq should be run with with one of the following options:
 
 * [force=TEXT](https://github.com/AdnanHodzic/auto-cpufreq/#overriding-governor)
     - Force use of either the powersave or performance governor. Setting to 'reset' goes back to normal mode
-
 * state
     - Show if the governor has been overriden with --force. Returns 'default' if there's no override
 
@@ -248,7 +247,8 @@ Necessary changes are temporarily made to the system which are lost with system 
 
 `sudo auto-cpufreq --force=governor`
 
-The system will be forced to use the governor of choice. The only governor options are "powersave" and "performance". Setting force to "reset" will disable any override. Please note that any set override will persist even after reboot.
+Force use of either "powersave" or "performance" governors. Setting to "reset" will go back to normal mode
+Please note that any set override will persist even after reboot.
 
 ### Install - auto-cpufreq daemon
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ auto-cpufreq is looking for [co-maintainers & open source developers to help sha
 * [auto-cpufreq modes and options](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-modes-and-options)
     * [monitor](https://github.com/AdnanHodzic/auto-cpufreq/#monitor)
     * [live](https://github.com/AdnanHodzic/auto-cpufreq/#live)
+    * [overriding governor](https://github.com/AdnanHodzic/auto-cpufreq/#overriding-governor)
     * [Install - auto-cpufreq daemon](https://github.com/AdnanHodzic/auto-cpufreq/#install---auto-cpufreq-daemon)
     * [Remove - auto-cpufreq daemon](https://github.com/AdnanHodzic/auto-cpufreq/#remove---auto-cpufreq-daemon)
     * [stats](https://github.com/AdnanHodzic/auto-cpufreq/#stats)
@@ -206,7 +207,7 @@ auto-cpufreq should be run with with one of the following options:
 * [stats](https://github.com/AdnanHodzic/auto-cpufreq/#stats)
     - View live stats of CPU optimizations made by daemon
 
-* force=TEXT
+* [force=TEXT](https://github.com/AdnanHodzic/auto-cpufreq/#overriding-governor)
     - Force use of either the powersave or performance governor. Setting to 'reset' goes back to normal mode
 
 * state
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ auto-cpufreq should be run with with one of the following options:
     - View live stats of CPU optimizations made by daemon
 
 * [force=TEXT](https://github.com/AdnanHodzic/auto-cpufreq/#overriding-governor)
-    - Force use of either the powersave or performance governor. Setting to 'reset' goes back to normal mode
+    - Force use of either the "powersave" or "performance" governor. Setting to "reset" goes back to normal mode
 
 * config=TEXT
     - Use config file at defined path
--- a/README.md
+++ b/README.md
@@ -210,8 +210,6 @@ auto-cpufreq should be run with with one of the following options:
 
 * [force=TEXT](https://github.com/AdnanHodzic/auto-cpufreq/#overriding-governor)
     - Force use of either the powersave or performance governor. Setting to 'reset' goes back to normal mode
-* state
-    - Show if the governor has been overriden with --force. Returns 'default' if there's no override
 
 * config=TEXT
     - Use config file at defined path
--- a/README.md
+++ b/README.md
@@ -206,6 +206,9 @@ auto-cpufreq should be run with with one of the following options:
 * [stats](https://github.com/AdnanHodzic/auto-cpufreq/#stats)
     - View live stats of CPU optimizations made by daemon
 
+* force TEXT
+    - Force use of either the powersave or performance governor. Setting to 'reset' goes back to normal mode
+
 * config TEXT
     - Use config file at defined path
 
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -90,16 +90,19 @@ def get_override():
     if os.path.isfile(STORE):
         with open(STORE, "rb") as store:
             return pickle.load(store)
+    else:
+        return "default"
 
-def set_override(override=False):
+def set_override(override):
     if override in ["powersave", "performance"]:
         with open(STORE, "wb") as store:
             pickle.dump(override, store)
         print(f"Set governor override to {override}")
     elif override == "reset":
-        os.remove(STORE)
+        if os.path.isfile(STORE):
+            os.remove(STORE)
         print("Governor override removed")
-    else:
+    elif override is not None:
         print("Invalid option.\nUse force=performance, force=powersave, or force=reset")
 
 
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -337,14 +337,6 @@ def footer(l=79):
     print("\n" + "-" * l + "\n")
 
 
-def daemon_not_found():
-    print("\n" + "-" * 32 + " Daemon check " + "-" * 33 + "\n")
-    print(
-        "ERROR:\n\nDaemon not enabled, must run install first, i.e: \nsudo auto-cpufreq --install"
-    )
-    footer()
-
-
 def deploy_complete_msg():
     print("\n" + "-" * 17 + " auto-cpufreq daemon installed and running " + "-" * 17 + "\n")
     print("To view live stats, run:\nauto-cpufreq --stats")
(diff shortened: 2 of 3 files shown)


AdnanHodzic:
> I noticed a couple of people had requested this and I wanted the feature myself, so I took the initiative. 

Awesome, always glad to hear this. I love it but I also have few remarks:

1. I can't get it to work, or am I not using the flag properly? 

```
sudo auto-cpufreq --force performance
auto-cpufreq: wrong invocation. try --help for help.
sudo auto-cpufreq --force "performance"
auto-cpufreq: wrong invocation. try --help for help.
sudo auto-cpufreq --force powersave    
auto-cpufreq: wrong invocation. try --help for help.
```

2. Also, I suggest to first have a check to see if daemon is running before this flag can be run. Take a [look in code](https://github.com/AdnanHodzic/auto-cpufreq/blob/122dd0f96c387fccb0555080bd84ccd7e0b796a0/auto_cpufreq/core.py#L1191). As you'll get this output, if you try to run `sudo auto-cpufreq --install` after you already ran `sudo auto-cpufreq --install`.

So in case flag was used and daemon is not running (--install) not run first, it would say something like: "Make sure auto-cpufreq --install  is run first (running as daemon) before this flag can be used"

3. Also as part of the flag description:

```
  --force TEXT           Force use of either powersave or performance
                         governors. Setting to 'reset' will go back to normal
                         mode
```

Could you please add "performance" and "powersave" in double quotes and "reset" in double quotes? Think it helps visually and is consistent with "--install_performance" flag description.

4. Also could we make it to be more verbose when `auto-cpufreq --state` is run, for example:

```
sudo auto-cpufreq --state
governor override set to "performance"
```

Thanks and looking forward to your reply

shadeyg56:
Hi, thanks for your reply

The option should be used as follows
`sudo auto-cpufreq --force=performance`
Maybe this should be made more clear in the README?

Also I think checking for the daemon running is a good idea. I'll update this PR as soon as I can with that change, as well as updating the flag description

as far as the output of `--state` goes,  in issue #467, the user wanted the `--state` flag to output something that could be used in a bar which is why I had it simply output the state with no verbosity. Maybe there could be a verbose option?

Thanks again.

AdnanHodzic:
> The option should be used as follows
> `sudo auto-cpufreq --force=performance`
> Maybe this should be made more clear in the README?

Yes, I was about to mention this as part of my original comment, let's [update this section of README](https://github.com/AdnanHodzic/auto-cpufreq/#how-to-run-auto-cpufreq).

> Also I think checking for the daemon running is a good idea. I'll update this PR as soon as I can with that change, 
> as well as updating the flag description

Awesome, looking forward to the change and take your time :)

> as far as the output of --state goes, in issue https://github.com/AdnanHodzic/auto-cpufreq/issues/467, 
> the user wanted the --state flag to output something that could be used in a bar which is why I had it 
> simply output the state with no verbosity. Maybe there could be a verbose option?

Ack, well now that this PR is here I would love to hear from @niksingh710 what he meant by:
> auto-cpufreq --state # (should return powersave,performance,reseted) So that the user can create a indicator in it's bar config.

In particular: 
> So that the user can create a indicator in it's bar config.

In #312 I mentioned that it would be great if auto-cpufreq had indicator/gnome-extension type of thing in the future, so would love to hear what user had in mind exactly here. 

Adding another "state-verbose" flag might be too much, so we might as well just explain it as part of README ...

niksingh710:
I was saying that indicator one just because of some users who use window managers and some kind of bar to show status and all.


E.g I use waybar so having a cli way to get the current state of autocpu-freq will be very easy to display there.

shadeyg56:
Maybe I'll change `--state` to something like `--bar`? And explain its function in readme and the `--help` flag

niksingh710:
Imo `--get-info` displaying information will be nice.
But `--bar` will do the job too.

AdnanHodzic:
@niksingh710 

> E.g I use waybar so having a cli way to get the current state of autocpu-freq will be very easy to display there.

Let me know what you end up with, full fledged auto-cpufreq running in tray or gnome extension would be great, but this could be a great beginning as well. Might even be worthy mentioning/showcasing on Readme?

> Imo --get-info displaying information will be nice.
> But --bar will do the job too.

I have an even better proposal: `--get-state`. To also avoid having too many flags you can set it to not be shown when you run `--help` @shadeyg56 [reference `--daemon`](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/bin/auto-cpufreq#L38), but should be mentioned and explained as part of Readme.

niksingh710:
> @niksingh710 
> 
> > E.g I use waybar so having a cli way to get the current state of autocpu-freq will be very easy to display there.
> 
> Let me know what you end up with, full fledged auto-cpufreq running in tray or gnome extension would be great, but this could be a great beginning as well. Might even be worthy mentioning/showcasing on Readme?
> 
> > Imo --get-info displaying information will be nice.
> > But --bar will do the job too.
>

In tray will be a better option as this will enable it for KDE or other DE those used system tray in their panel.

shadeyg56:
I think tray should he handled in a separate PR to keep this one from having too much, but it's a good idea. I'll focus on the flag for now

AdnanHodzic:
> In tray will be a better option as this will enable it for KDE or other DE those used system tray in their panel.
...
> I think tray should he handled in a separate PR to keep this one from having too much, but it's a good idea. I'll focus on the flag > for now

Sounds good!

shadeyg56:
Ok new commits have been pushed.

FIrst off, in commit 573452a69615d8fc8e9ebcd905e188c8ed975449, I added two functions for sending a message when a user tries to use `--force` before running `--install`. Also if an override is set and a user calls `--remove`, it will now remove the override. I also renamed `remove()` to `remove_daemon()`. I will come back to this

Now in 08e5a2bd78f67456925442d0d5826f5fa475d6c9, you can see I changed the `--install/--remove` decorator. This is because this is not the intended use of boolean flags which can be found in the [Click docs](https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags). Now there was also a function named `remove()` in core.py. This was conflicting with the `remove` variable so I had to change the name. This improves overall readability anyways. I also renamed `--state` to `--get-state` for now.

Finally, the last commit is just the README update. I made it a little clearer and also added a dedicated section for overriding governors and hyperlinked to it.

I'd love to hear your feedback. On another note, I'm willing to do another PR for the tray menu. Just know I've never done that before, so it will be a learning experience, but I am down to do it.

Sorry for the long message haha

AdnanHodzic:
> I'd love to hear your feedback. On another note, I'm willing to do another PR for the tray menu. 
> Just know I've never done that before, so it will be a learning experience, but I am down to do it.

That that would be great!

> Sorry for the long message haha

No worries, I actually prefer when folks are more verbose in their PR's as it reduces chances of having any ambiguity. Especially since my reply won't be any shorter :D 

Everything works as expected now, but I do have few remarks.

**Observations no: 1:** 

If I run `auto-cpufreq --get-state` without running `auto-cpufreq --install` it will still just return "default", where I think it'd be better to return output of `no_stats_msg` func, i.e:

`ERROR: auto-cpufreq stats are missing.\n\nMake sure to run: "auto-cpufreq --install" first'`

Because, it's already outputting state, and yet auto-cpufreq daemon hasn't even been installed yet.

I also noticed, that although `daemon` flag is hidden, if you run `sudo auto-cpufreq --daemon` it'll start executing:

```
sudo auto-cpufreq --daemon

Using settings defined in /etc/auto-cpufreq.conf file
```

Where it would also be good idea to return similar message, i.e: 

```
 'ERROR: auto-cpufreq daemon is missing.\n\nMake sure to run: "auto-cpufreq --install" first'
```

So I think it would be a good idea to rename `no_stats_msg` func to `no_daemon_msg` and return its contents in more generic fashion: 

` 'ERROR: auto-cpufreq daemon is missing.\n\nMake sure to run: "auto-cpufreq --install" first'` 

Then this way, it could be use for both cases of `--state` and `--daemon` if it's run without --install first?

**Observation no 2:**

Could you still do this as I mentioned in my initial comment:

```
Also as part of the flag description:
  --force TEXT           Force use of either powersave or performance
                         governors. Setting to 'reset' will go back to normal
                         mode
```

Could you please add "performance" and "powersave" in double quotes and "reset" in double quotes? Think it helps visually and is consistent with "--install_performance" flag description.

**Observation no 3:** 

I just realized now, it would be good if overwrite state (output of `auto-cpufreq --get-state`) would be displayed in ``auto-cpufreq --stats`. For example bottom portion looks like:

```

Total CPU usage: 1.8 %
Total system load: 0.77
Average temp. of all cores: 48.25 °C 
```

I think it would be a good idea to be like:

````

Warning: governor overwritten using `--force` flag.

Total CPU usage: 1.8 %
Total system load: 0.77
Average temp. of all cores: 48.25 °C 
```

This way we could even hide `--get-state` by default, and yet it could still be used as part of tray icon.

Looking forward to hearing your thoughts on this.

shadeyg56:
1. Sounds good to me. Cuts down on some functions

2. Ah I knew I was missing something in the last commits. Thanks for reminding me.

3. Sure sounds like a good idea.

I'll update this PR once again when I've made the changes

shadeyg56:
Alright, commits have been merged.  Just a few quick comments

1. 
> So I think it would be a good idea to rename no_stats_msg func to no_daemon_msg and return its contents in more generic fashion:
 'ERROR: auto-cpufreq daemon is missing.\n\nMake sure to run: "auto-cpufreq --install" first'
Then this way, it could be use for both cases of --state and --daemon if it's run without --install first?

So I did make this the case for both `--get-state` and `--stats`. The error messages have been cut down in commit a339b82. I did not make this true for `--daemon` because using `sudo auto-cpufreq --daemon` is technically running the daemon so you cannot check for it here without checking that the service files have been installed. 
In my opinion, if you do not want the user to be able to run this flag themselves, then the `--daemon` flag should be removed altogether, and a separate file should be created for running the daemon. Then the service files should be updated to run that separate daemon script. This should probably be handled in a different PR.

2. README should be updated with new changes

3. `--stats` will now show a governor override warning

That should be all. Let me know what you think :)

AdnanHodzic:
1. Sounds good! 

2. Small thing, I still see double quotes missing around "powersave", "performance" in Readme as they are now in flag description :) 

3. Perfect

> That should be all. Let me know what you think :)

One final thing before I merge this with `master` branch. I noticed that if I I use `sudo auto-cpufreq --force=performance`, I'll get this (as expected):

```
Warning: governor overwritten using `--force` flag.
```

After which if I do `sudo auto-cpufreq --force=reset `, and no matter how low the load goes, `Setting to use: "performance" governor` never changes, unless I switch it to `sudo auto-cpufreq --force=powersave`, then it'll switch it back to `powersave`.

While overwriting governor with `powersave`, if I run i.e: `stress --cpu 8 --io 4 --vm 4 --timeout 15` it'll switch back to `performance` to do the heavy lifting (as it should). But I just never see it go back to other governor when overwritten with `performance`.  I'll keep testing, but do you have this on your side?

Because, same thing will happen with `--config` where changes won't be updated whatever was changed in config file, until auto-cpufreq process was restarted (but this was never implemented).

Sidenote, after these changes are removed, and I make sure it all works great as part of snap package as well I'll probably proceed to remove `--install_performance` flag and all its functionality  as this flag is lot more elegant way to go about this.

shadeyg56:
Oh my goodness I could've swore I put the quotes in the README 😂. I guess I only did it for the dedicated section. 

> While overwriting governor with powersave, if I run i.e: stress --cpu 8 --io 4 --vm 4 --timeout 15 it'll switch back to performance to do the heavy lifting (as it should) and go back to powersave once done. But I just never see it go back to other governor when overwritten with performance. I'll keep testing, but do you have this on your side?

This shouldn't be happening. The whole point of the `--force` flag is to use a governor of your choice all the time so it should never switch itself if an override is active. I'll check this out later and see if I can reproduce/fix this.

shadeyg56:
> After which if I do sudo auto-cpufreq --force=reset , and no matter how low the load goes, Setting to use: "performance" governor never changes, unless I switch it to sudo auto-cpufreq --force=powersave, then it'll switch it back to powersave.
While overwriting governor with powersave, if I run i.e: stress --cpu 8 --io 4 --vm 4 --timeout 15 it'll switch back to performance to do the heavy lifting (as it should) and go back to powersave once done. But I just never see it go back to other governor when overwritten with performance. I'll keep testing, but do you have this on your side?

I was unable to reproduce this. On my system, it works as intended. Using `--force` sets auto-cpufreq to use that governor all the time until `--force=reset` is called

tpongo-afk:
Hello everyone, I was reading through this PR and noticed that you only mention powersave and performance mode. Due to having the amd_pstate driver and issue #398 I use the conservative governor when on battery. Do you have in mind including the other [governors](https://wiki.archlinux.org/title/CPU_frequency_scaling#Scaling_governors)?

shadeyg56:
Hey @tpongo-afk. This is actually already the case. The governor override will use your config file as well

So if I do `sudo auto-cpufreq --force=powersave` and I have my config file set to use conservative on battery, then it will override to use conservative. This is bad naming on my part. Perhaps the force flag should be changed to `sudo auto-cpufreq --force=battery` and  `sudo auto-cpufreq --force=charger`?

I'm not sure what the best way to do it is as I feel "battery" and "charger" are also a little confusing as to what the flag is supposed to do. Maybe it should just be up to the user to be able to use any of the available governors? Let me know what you think.

AdnanHodzic:
@shadeyg56 

> I was unable to reproduce this. On my system, it works as intended. Using `--force` sets auto-cpufreq to use that governor all the time until `--force=reset` is called

Good, and what about: 

> After which if I do sudo auto-cpufreq --force=reset , and no matter how low the load goes, Setting to use: "performance" governor never changes, unless I switch it to sudo auto-cpufreq --force=powersave, then it'll switch it back to powersave.

Can you reproduce this? As this is more of a potential issue.

> In the meantime, I have updated the README... hopefully for the final time lol

No worries, that's why we have these code reviews :) 

> So if I do sudo auto-cpufreq --force=powersave and I have my config file set to use conservative on battery, then it will override to use conservative. This is bad naming on my part. Perhaps the force flag should be changed to sudo auto-cpufreq --force=battery and sudo auto-cpufreq --force=charger?

I suggest to simply mention this as part of a Readme, using `--force=charger/battery` is just not right.

shadeyg56:
> After which if I do sudo auto-cpufreq --force=reset , and no matter how low the load goes, Setting to use: "performance" governor never changes to i.e "powersave", unless I switch it to sudo auto-cpufreq --force=powersave, then it'll switch it back to powersave.

Hm I guess I was confused by what you meant initially. I believe I tested this when I initially submitted the PR but I will double check

shadeyg56:
I am unable to reproduce it. I'd like to explain exactly what is happening on my system so there is no confusion.

I am running my laptop off of battery, so the "powersave" governor is in use.
I run `sudo auto-cpufreq --force=performance`
Now the "performance" governor is always in use.
I do whatever performance stuff I want to do
Then I run `sudo auto-cpufreq --force=reset` and it goes back to using "powersave" because my laptop is running off battery still

\he override runs all the time, until `--force=reset` is used. After it is used, auto-cpufreq goes back to using the governor it should be using based on the config file.

shadeyg56:
Awesome. Thanks for your correspondence. 

As far as the tray goes, I might open up a discussion or you can reach me on Discord at shadeyg56#2429 so we can discuss the details of the tray menu so I can know exactly what you have in mind :)

PR #501, opened 2023-04-27
updated the systemd service to update the script

boredsquirrel:
I guess this is not the perfect solution, also I am not sure if the environment interrupts the process. But in theory it should check for an update, and if there is one reinstall the program, then start it.

### DIFF
--- a/scripts/auto-cpufreq.service
+++ b/scripts/auto-cpufreq.service
@@ -7,7 +7,7 @@ Type=simple
 User=root
 WorkingDirectory=/opt/auto-cpufreq/venv
 Environment=PYTHONPATH=/opt/auto-cpufreq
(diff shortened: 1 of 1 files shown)


meator:
Services shouldn't be responsible for this; see https://github.com/AdnanHodzic/auto-cpufreq/pull/500#pullrequestreview-1891195465

AdnanHodzic:
> Services shouldn't be responsible for this; see [#500 (review)](https://github.com/AdnanHodzic/auto-cpufreq/pull/500#pullrequestreview-1891195465)

I agree with what was said here, which leads me to a question why not extend the `--update` and improve this process?

AdnanHodzic:
Closed due to inactivity.

Issue #498, opened 2023-04-12
Incorrect temperatures


pimpmyname2:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted?
Yes.
 
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)?
No. Installation succeeded.

- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)?
Yes.
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)?
Yes.

### Error output:
```text
No errors.
```
---

### System information:

Add/paste output of:

```
Using settings defined in /etc/auto-cpufreq.conf file

-------------------------------------------------------------------------------

Linux distro: Ubuntu 22.04 Jammy Jellyfish
Linux kernel: 5.19.0-38-generic
Processor: AMD Ryzen Threadripper 2950X 16-Core Processor
Cores: 32
Architecture: x86_64
Driver: acpi-cpufreq

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 3500 MHz
CPU min frequency: 2800 MHz

Core    Usage   Temperature     Frequency
CPU0:     0.0%     26 °C     2226 MHz
CPU1:     3.0%     26 °C     2224 MHz
CPU2:     1.0%     26 °C     2224 MHz
CPU3:     1.0%     26 °C     3122 MHz
CPU4:     1.0%     26 °C     2727 MHz
CPU5:     0.0%     26 °C     4098 MHz
CPU6:     0.0%     26 °C     2212 MHz
CPU7:     1.0%     26 °C     2704 MHz
CPU8:    11.0%     26 °C     2796 MHz
CPU9:     0.0%     26 °C     2800 MHz
CPU10:    8.0%     26 °C     2794 MHz
CPU11:    4.0%     26 °C     2794 MHz
CPU12:    1.0%     26 °C     2793 MHz
CPU13:    0.0%     26 °C     2800 MHz
CPU14:    1.0%     26 °C     2800 MHz
CPU15:    4.0%     26 °C     2794 MHz
[...]

PR #513, opened 2023-05-19
Check for CPU label if no coretemps

marchwicki:
If the individual core temps are not available this will attempt to use the sensor labeled "CPU" before falling back to the first sensor in the dictionary. The first sensor in my case was not the CPU.  

This relates to this issue: https://github.com/AdnanHodzic/auto-cpufreq/issues/498


AdnanHodzic:
LGTM, thank you for your contribution, you'll be given credit for your work as part of upcoming release.

PR #533, opened 2023-06-30
cpufreqctl: don't try writing to non-existent files

cmm:
When you give "nosmt" to your kernel, you still get directories under /sys that correspond to inactive hyperthreads, but there's nothing writable inside those directories, which causes auto-cpufreq to generate lots of logspam.

### DIFF
--- a/scripts/cpufreqctl.sh
+++ b/scripts/cpufreqctl.sh
@@ -100,6 +100,12 @@ function driver () {
   cat $FLROOT/cpu0/cpufreq/scaling_driver
 }
 
+function write_value () {
+  if [ -w $FLNM ]; then
+    echo $VALUE > $FLNM
+  fi
+}
+
 function set_driver () {
   DRIVER=`driver`
   case $DRIVER in
@@ -136,7 +142,7 @@ function set_governor () {
     while [ $i -ne $cpucount ]
     do
       FLNM="$FLROOT/cpu"$i"/cpufreq/scaling_governor"
-      echo $VALUE > $FLNM
+      write_value
       i=`expr $i + 1`
     done
   else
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Nice catch, not sure how this was missed until now. Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

cmm:
> you will be credited for your work

I would rather not be, if it's all the same to you, having taken a closer look at the sources.

> not sure how this was missed until now

Oh man.   Many, many possible reasons.  Where to even begin.

(I apologize if this comment sounds hostile.  I do not mean it to be!  At this point I choose to believe that this whole project is an art performance, meant to show what's wrong with Linux user-space ecosystem in general and also how not to write software.  Right?)

AdnanHodzic:
> > you will be credited for your work
> 
> I would rather not be

Noted, you will not be credited. 

> if it's all the same to you, having taken a closer look at the sources.

Not sure what you mean by this.

> (I apologize if this comment sounds hostile. I do not mean it to! At this point I choose to believe that this whole project is an art performance, meant to show what's wrong with Linux user-space ecosystem in general and also how not to write software. Right?)

> I apologize if this comment sounds hostile. I do not mean it to! 

Thanks for the heads up, because what follows definitely made it one of the most hostile and ill mannered comments :) 

> At this point I choose to believe that this whole project is an art performance, meant to show what's wrong with Linux user-space ecosystem in general and also how not to write software. Right?)

Needless to say, hours and hours of various contributors work, 4k stars on Github, 61 contributors, various mentions in reputable sites and news paper definitely don't agree with this statement. 

If you think something is wrong, feel free to improve & fix it. Otherwise you're also free to stay away from this project.

PR #526, opened 2023-06-20
Update feature added

cosmos1721:
As discussed in issue #341 the commit changes are hereby elaborately described:
- update using `sudo ./auto-cpufreq-installer` and `sudo auto-cpufreq --update`
- version number will update automatically everytime the package is updated to the latest release
- some functions that have not been defined (showing warnings) have been commented and bypassed 
- when updated using installer:
       -- it removes the package and clones with latest release
- when updated using `--update` mode, the package can be found in $HOME directory 
- README.md updated
 

Please let me know if changes are required.

PR #535, opened 2023-07-13
update feature

cosmos1721:
update feature in continuation with closed PR #526  and issue #341


snap/snapcraft.yaml
```diff
@@ -39,11 +40,15 @@ parts:
       snapdaemon.sh: usr/bin/snapdaemon
 
 plugs:
-   etc-auto-cpufreq-conf:
+  network:
```
AdnanHodzic, line 43: Not sure why this part was added, but lines 43 to 46 should be removed. As upload of this build of this snap with these changes will otherwise fail, i.e:

```
snapcraft upload auto-cpufreq_1.9.9_amd64.snap --release beta
Issues while processing snap:g                                                                                                                                       
- unknown attribute 'allow-auto-connection' for interface 'network' (plugs)    
```

Hence this whole code block should look like just as it does on `master`:

```
plugs:
   etc-auto-cpufreq-conf:
    interface: system-files
    write:
    - /etc/auto-cpufreq.conf
```

AdnanHodzic:
Looks great, this is what I wanted #526 to look like :slightly_smiling_face: 

Before I merge this there's one minor change that needs to be done. Added as a comment for [snap/snapcraft.yaml](https://github.com/AdnanHodzic/auto-cpufreq/pull/535/files#r1263342097)

snap/snapcraft.yaml
cosmos1721: it was used to give network access to the container, but its not required now

cosmos1721: made changes according to it now

snap/snapcraft.yaml
```diff
@@ -39,11 +40,11 @@ parts:
       snapdaemon.sh: usr/bin/snapdaemon
 
 plugs:
-   etc-auto-cpufreq-conf:
+  etc-auto-cpufreq-conf:
```
AdnanHodzic, line 43: I asked this already mentioned multiple times #526, please do not add any characters that are needed. Please remove the empty white space on this line. Depending on file extension this can even break things.

README.md
```diff
@@ -99,12 +101,22 @@ Get source code, run installer and follow on screen instructions:
 git clone https://github.com/AdnanHodzic/auto-cpufreq.git
 cd auto-cpufreq && sudo ./auto-cpufreq-installer
 ```
+### Update using auto-cpufreq-installer
+* The feature is available from version *1.9.8*. If your current version is below the same, Get source code, run installer and follow on screen instructions:
```
AdnanHodzic, line 105: I would remove everything after "* The feature is available from version *1.9.8*."

```
If your current version is below the same, Get source code, run installer and follow on screen instructions:
...
```

until line 113. Because it just extends the README for no reason and it might break things for some people, in case they are running on snap or AUR or just have auto-cpufreq under different path. If they want instructions just refer them to Install seciton.

README.md
```diff
@@ -120,11 +132,11 @@ auto-cpufreq makes all decisions automatically based on various factors like cpu
 
 ### 1: power_helper.py script (Snap package install **only**)
 
-When installing auto-cpufreq using [auto-cpufreq-installer](#auto-cpufreq-installer) if it detects [GNOME Power profiles service](https://twitter.com/fooctrl/status/1467469508373884933) is running it will automatically disable it. Otherwise this daemon will cause conflicts and various other performance issues.
+When installing auto-cpufreq using [auto-cpufreq-installer](#auto-cpufreq-installer) if it detects [GNOME Power profiles service](https://twitter.com/fooctrl/status/1467469508373884933) is running it will automatically disable it. Otherwise this daemon will cause conflicts and various other performance issues. 
 
 However, when auto-cpufreq is installed as Snap package it's running as part of a container with limited permissions to your host machine, hence it's *highly recommended* you disable GNOME Power Profiles Daemon using `power_helper.py` script.
 
-**Please Note:**<br>
+**Please Note:**
```
AdnanHodzic, line 139: As I said in other comments, please do not make changes that are necessary.

Or maybe you have a reason why `**Please Note:**<br>` was removed? As Missing `<br`>` adds a new line which I think is needed.

snap/snapcraft.yaml
cosmos1721: done

README.md
cosmos1721: done

README.md
cosmos1721: i might have accidentally removed it
altered, done

AdnanHodzic: Did you push the changes because I still see it how it was?

snap/snapcraft.yaml
AdnanHodzic: Did you push the changes because I still see it how it was?

AdnanHodzic:
LGTM! Thank you for your contribution, and your perseverance is admired in this process. :slightly_smiling_face: 

You will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases) and your future contributors are more then welcome!

AdnanHodzic:
@cosmos1721 changes are already merged and I didn't want to mention it as part of initial PR. But I already have a few suggestions how to improve this functionality: 

It would be great idea to give users an option to chose where cloned auto-cpufreq dir wil be placed.

I.e: `sudo auto-cpufreq --update --dir /code` will place it to `/home/$user/code/auto-cpufreq`

As I already see people getting annoyed by this. 

It's also a good idea to [update README section ](https://github.com/AdnanHodzic/auto-cpufreq#update-using-auto-cpufreq-installer) will info where will auto-cpufreq code be cloned by default as currently your regular user will have no idea about this. 

Also it would be great idea to [update README section ](https://github.com/AdnanHodzic/auto-cpufreq#update-using-auto-cpufreq-installer) with info about [repository structure](https://github.com/AdnanHodzic/auto-cpufreq/blob/3dffa23e0007e99577fded98eac77c82a3b49eea/auto-cpufreq-installer#L262). Although ideally this would be tracked through code.

cosmos1721:
sure, thats a great option to work on.
will add the required info to README section

cosmos1721:
> LGTM! Thank you for your contribution, and your perseverance is admired in this process. slightly_smiling_face
> 
> You will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases) and your future contributors are more then welcome!

thanks a lot

cosmos1721:
> Related to this instead of using [current working directory of user](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto_cpufreq/core.py#L188C4-L188C4) let's just add this to `/opt/auto-cpufreq/source` so it will always be there and there won't be changes to dir structure. Especially since install location `/opt/auto-cpufreq/venv` is also under `/opt/auto-cpufreq`.

what I'm thinking is to give the user an option in which
if the user only uses ```auto-cpufreq --update```, it will save it to the provided location,but if the user adds --dir, it can work too

AdnanHodzic:
> what I'm thinking is to give the user an option in which
> if the user only uses `auto-cpufreq --update`, it will save it to the provided location,but if the user adds --dir, it can work too

Yes, this is what I also meant i[n updated comment above](https://github.com/AdnanHodzic/auto-cpufreq/pull/535#issuecomment-1636877526). But, if user doesn't specify anything, let's make it so by default it goes to `/opt/auto-cpufreq/source` (and update README with this info) instead of [current working user](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto_cpufreq/core.py#L188C4-L188C4) which in this case will always be `/root/auto-cpufreq` since `--update` flag will have to be run with `sudo`.

PR #538, opened 2023-07-23
added optional path in --update

cosmos1721:
in continuation of PR #535 
as per your suggestions, an optional add to path has been added.
i.e. `auto-cpufreq --update=path/to/dir`

i was expecting to use:
 `auto-cpufreq --update path/to/dir`
but  there are some unknown issues


AdnanHodzic:
1. Let's set to use `/opt/auto-cpufreq/source` as the default path as part of `--update` as part of this PR, and then having ability to set "custom path" can be made as part of one of following iterations. 

Besides this I'll leave couple of comments for the code itself, as I found some bugs.

2. What's the proper mechanism for testing `--update` changes? As changing i.e `VERSION = "1.9.8"` line in setup.py had no impact, and even changing auto-cpufreq-installer file to trick it into that there are updates:

```
    # Fetch the latest release information from GitHub API
    #latest_release=$(curl -s "https://api.github.com/repos/$repository/releases/latest")
    # Extract the latest release version without using jq
    #latest_version=$(echo "$latest_release" | grep -o '"tag_name": "[^"]*' | cut -d'"' -f4)
    latest_version=v1.9.9 # same case with v2.0
    # Get the current version of auto-cpufreq
      #installed_version=$(pip list | awk '/auto-cpufreq/ {print $2}')
    installed_version=$(grep -oP "(?<=__requires__ = 'auto-cpufreq==)\d+(\.\d+)+" /opt/auto-cpufreq/venv/bin/auto-cpufreq)
```

kept returning: 

```
sudo auto-cpufreq --update 
auto-cpufreq is up to date
```

auto-cpufreq-installer
```diff
@@ -192,6 +192,8 @@ elif [ -f /etc/os-release ];then
 else # In case /etc/os-release doesn't exist
 	manual_install
 fi
+# to clone the repo
+mkdir /opt/auto-cpufreq/source
 }
```
AdnanHodzic, line 197: Let's only run this if `--update` is run, as otherwise when If `sudo ./auto-cpufreq-installer --install` is run, and `/opt/auto-cpufreq/source` was created before following error will pop up:

`mkdir: cannot create directory ‘/opt/auto-cpufreq/source’: File exists`

bin/auto-cpufreq
```diff
@@ -226,9 +226,22 @@ def main(config, daemon, debug, update, install, remove, live, log, monitor, sta
                 remove_complete_msg()
         elif update:
             root_check()
+            custom_dir = "/opt/auto-cpufreq/source"
+            for arg in sys.argv:
+                if arg.startswith("--update="):
+                    custom_dir = arg.split("=")[1]
+                    sys.argv.remove(arg)
+                    
+            if "--update" in sys.argv:
```
AdnanHodzic, line 235: I guess we don't need this part now if we'll use `custom_dir` with set location for now

bin/auto-cpufreq
```diff
@@ -238,19 +251,27 @@ def main(config, daemon, debug, update, install, remove, live, log, monitor, sta
                 verify_update()
                 ans = input ("Do you want to update auto-cpufreq to the latest release? [y/n]: ")
                 valid_options = ['y', 'Y', 'yes', 'YES', 'Yes']
+                if custom_dir != "/opt/auto-cpufreq/source":
```
AdnanHodzic, line 254: This part can also be removed with set custom path

auto-cpufreq-installer
AdnanHodzic: Let's move 

```
# to clone the repo
mkdir /opt/auto-cpufreq/source
```

after

```
elif update:
```

in `bin/auto-cpufreq` file, as this way `/opt/auto-cpufreq/source` won't be created until you run `auto-cpufreq --update`

You can even use do it using:

```
if not os.path.exists(custom_dir):
    os.mkdir(custom_dir)
```

Since custom_dir variable already exists.

AdnanHodzic: But now I see you're creating this as part line 254 of `bin/auto-cpufreq` file which makes these 2 lines obsolete?

So let's remove them

```
 to clone the repo
mkdir /opt/auto-cpufreq/source
```

bin/auto-cpufreq
```diff
@@ -238,19 +251,27 @@ def main(config, daemon, debug, update, install, remove, live, log, monitor, sta
...
+                    if not os.path.exists(custom_dir):
+                        input("Directory does not exist, press Enter to create and proceed or ctrl+c to exit")
```
AdnanHodzic, line 256: Let's remove this line  (256) and create directory without asking anything if it doesn't exist.

bin/auto-cpufreq
```diff
@@ -238,19 +251,27 @@ def main(config, daemon, debug, update, install, remove, live, log, monitor, sta
...
+                        os.makedirs(custom_dir)
+                else:
```
AdnanHodzic, line 258: Is logic from line 258 to 261 to delete the dir and its contents, so new contents can be cloned?

bin/auto-cpufreq
```diff
@@ -238,19 +251,27 @@ def main(config, daemon, debug, update, install, remove, live, log, monitor, sta
...
+                    if os.path.exists(custom_dir):
+                        os.rmdir(custom_dir)
+                        os.makedirs(custom_dir)
+                    else:
```
AdnanHodzic, line 262: I don't think we need line 262 & 263, as this directory would've been created above if it doesn't exist.

AdnanHodzic:
LGTM, thank you for your contribution!

PR #540, opened 2023-08-12
make it much better with improvements

max-privatevoid:


AdnanHodzic:
Since a lot of folks use auto-cpufreq package, instead of killing support for it completely I decided to make this process as graceful as possible in hopes for people to slowly start moving away from it.

In the latest commit, I've made Snap package [deprecation notice.](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store), and those who run GUI functionality will not be able to do so due to Snap package confinement limitations, and [they will be informed of that](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto_cpufreq/gui/app.py#L61) and same is with CLI.

Hence, thank you for your contribution, but I'm closing this PR for now. I might re-open in future and tag you if you really want to remove all references to Snap package since there's a lot of them.

PR #550, opened 2023-09-09
Fixes auto-cpufreq --update for Arch based distros for non AUR installations

rootCircle:
Details:
-  Updated logic now checks if auto-cpufreq is installed via AUR or not (pacman -Q auto-cpufreq). If found, it recommends using the system package manager to refresh it.

-  Removed the obsolete yay checking as yay is considered a Pacman wrapper. (Need confirmation from @cosmos1721 for intent) 

Fixes:
Persistant `Arch-based distribution with AUR support detected. Please refresh auto-cpufreq using your AUR helper.` error, even when you haven't installed auto-cpufreq from AUR, on Arch Distro.

References:
[Arch Wiki - AUR Helpers](https://wiki.archlinux.org/title/AUR_helpers)

### DIFF
--- a/bin/auto-cpufreq
+++ b/bin/auto-cpufreq
@@ -232,7 +232,7 @@ def main(config, daemon, debug, update, install, remove, live, log, monitor, sta
 
                 print("Please update using snap package manager, i.e: `sudo snap refresh auto-cpufreq`.")
                 #check for AUR 
-            elif subprocess.run(["bash", "-c", "command -v yay >/dev/null 2>&1"]).returncode == 0 or subprocess.run(["bash", "-c", "command -v pacman >/dev/null 2>&1"]).returncode == 0:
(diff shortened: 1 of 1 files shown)


shadeyg56:
Yeah, `yay` should never have been used in the first place as a lot of Arch systems don't even use it, while `pacman` is available on all Arch systems

cosmos1721:
> * Removed the obsolete yay checking as yay is considered a Pacman wrapper.

pls do, can you also look up other configurations that are done for Arch?

AdnanHodzic:
Great one @rootCircle! Because I was proposing the same to @cosmos1721 as part of of the comments in original PR. But idea is that we'd wait for a PR like this one coming from Arch user :)

rootCircle:
Updated to v2.0.0, update seems to work fine on Arch based distros!

Congrats for the v2.x.x feat! :tada:

PR #551, opened 2023-09-10
replaced #!/bin/bash  with  #!/usr/bin/env bash

robertosw:
Some Distros do not have /bin/bash by default, but have bash installed at some other location and in PATH.
To allow all of these distros to use these shell scripts, change the shebangs to search for bash in PATH.

More to why this is better [here](https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my)

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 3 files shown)


AdnanHodzic:

PR #554, opened 2023-09-13
GUI: finishing touches and bug fixes

shadeyg56:
As far as I can tell, we have all the changes made for the GUI for it to be released. Not much to say here, Adnan knows what we have been discussing and the changes that have been made are available in the various commits.

Let me know if I missed anything!


AdnanHodzic:
Amazing work and great job as always!

PR #567, opened 2023-10-01
Update auto-cpufreq-installer

p3nguin-kun:
- Auto install packages on RedHat-based


AdnanHodzic:
LGTM, just a question before I merge this, are there any drawbacks to switching to yum is dnf omnipresent/default on RedHat based distros nowdays?

PR #572, opened 2023-10-03
nix: replace local derivation with nixpkgs derivation

shadeyg56:
Pretty simple update to the Nix flake here. 
Basically when I first created the flake `setuptools-git-versioning` was not available in nixpkgs so I included the package myself. Somebody submitted it to nixpkgs pretty recently, so I deleted the local package and swapped it for the one from the nixpkgs repo.


AdnanHodzic:
LGTM!

PR #576, opened 2023-10-10
feat: support for pyproject.toml

rootCircle:
### Changes
- Support for pyproject as per [PEP-518](https://peps.python.org/pep-0518/)
- setup.py renamed to setup.old.py 

### Needs help in checking if
- Snap builds works
- Nix build works

### Couldn't Implement
- [ ] python setup.py install  ***--record*** equivalent in this case 

### Breaking Changes
- Dynamic versioning is not the same as the one provided by pkg_marker, however appropriate apis in core.py has been provided in to fix that problem
  eg: 
          Old Version Scheme : auto-cpufreq==2.0.0-e7dyadda
           New Version SCheme : auto-cpufreq-2.0.0-e7dyadda
   `auto-cpufreq --version` won't be affected by this however.

- bin files have moved from `bin/<exec>` to `auto-cpufreq/bin/<exec>.py`

### Tested
- [x] Installation works
- [x] auto-cpufreq --install works 
- [x] GUI works

## Running the project

- ```bash
  git clone git@github:AdnanHodzic/auto-cpufreq.git
  cd auto-cpufreq
  poetry install
  poetry run auto-cpufreq --help
  ```

- Alternatively, we can use editable pip install for development purposes
  ```bash
  git clone git@github:AdnanHodzic/auto-cpufreq.git
  cd auto-cpufreq
  # Setup  your Virtual Environment
  pip3 install -e .
  auto-cpufreq
  ```


shadeyg56:
Awesome, I can test to see if it builds on Nix. Is there any point into hanging onto the setup.py?

shadeyg56:
The Nix package needed some slight modifications to work, but I have pushed the changes to your branch to make it build on Nix!

rootCircle:
> The Nix package needed some slight modifications to work, but I have pushed the changes to your branch to make it build on Nix!

Sounds great! Thanks :smile:

AdnanHodzic:
LGTM & thank you for your contibution! 

Wondering if "Running the project" part should've been included anywhere into README? Also let's keep an eye if `pip3 install .` is not working on plethora of distros that auto-cpufreq that can be installed on nowdays :)

rootCircle:
> Wondering if "Running the project" part should've been included anywhere into README?

Sure, I can

> Also let's keep an eye if pip3 install . is not working on plethora of distros that auto-cpufreq that can be installed on nowdays :)

How about replacing that with `pip` as fallback, if that doesn't works out?

@AdnanHodzic I wanted to tell that if this update breaks few things due to record `files.txt` in tool_remove in `auto-cpufreq-installer`, although I have tested if installation and removal things works or not!

PR #581, opened 2023-10-13
chore(auto-cpufreq-installer): removed dependence on record files for removal 

rootCircle:
As mentioned in #576, pyproject doesn't support record files, so this PR removes those dependence entirely.

## Minor Changes also include

- Docs for installing project using poetry for developers
- Removed setup.py and requirements.txt


AdnanHodzic:
LGTM! Thank you for your contribution

PR #582, opened 2023-10-13
refactor(flake): improve usage

isabelroses:
changes:
- allow for different system architectures
- separate dev deps into separate file
- gui now functions properly when installing though flake

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,21 +1,20 @@
 {
-    inputs = {
-        nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
-    };
+  inputs = {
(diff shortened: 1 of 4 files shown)


shadeyg56:
Hi, I am the original author of the flake. I just started using Nix about 2 months ago so my Nix writing skills are not the greatest haha so I appreciate this

I hadn't noticed the GUI wasn't working since I had the files I needed in `/usr/local` from messing around with `auto-cpufreq` before having the flake so thanks for catching that!

I did push a few changes. I removed darwin from being a build system because `auto-cpufreq` only supports the Linux kernel. Also the icon still wasn't working in the about dialog so I added that to the substitute as well.

The module was still using `x86_64-linux` so I updated it to inherit the system.

Thanks for your contribution and you'll receive credit in the next release!

PR #587, opened 2023-10-25
Feature: Allow GUI to detect available update and install it

shadeyg56:
20231025_15h35m12s_grim

I brought this up in the Discord and started working on it. 

Changes
- renamed `verify_update` to `check_for_update` and will now return True if an update was found
- GUI will prompt for an update if `check_for_update()` returns True
- Calls `auto-cpufreq --update` in the background

In order to test this, you need to edit the version to something less than 2.0.0 in `pyproject.toml` and reinstall or update what the version function returns in `core.py`


AdnanHodzic:
Code LGTM, and I've verified that it works as expected. Amazing work and as always thank you for your contribution!

Issue #145, opened 2021-01-12
Ryzen 4750U not showing temperature output with no error


JJGadgets:
Hey, my CPU temperature doesn't seem to be reported on auto-cpufreq on my Ryzen 7 4750U ThinkPad T14 on Manjaro KDE kernel 5.9.16-1. s-tui seems to work just fine tho, so does ZenMonitor. I do not get an error in auto-cpufreq despite the temperature clearly being not displayed, compared to #14.

### Error output:
```text
---------------------------- CPU frequency scaling ----------------------------

Battery is: charging

Setting to use: "performance" governor

Total CPU usage: 16.4 %
Total system load: 3.48
Average temp. of all cores: nan °C
```
---

### System information:
```text
-------------------------------------------------------------------------------

Linux distro: Manjaro Linux 20.2.1 Nibia
Linux kernel: 5.9.16-1-MANJARO
Processor: AMD Ryzen 7 PRO 4750U with Radeon Graphics
Cores: 16
Architecture: x86_64
Driver: acpi-cpufreq

------------------------------ Current CPU states ------------------------------

CPU max frequency: 1700 MHz
CPU min frequency: 1400 MHz

	 Usage  Temperature  Frequency
CPU0:	 11.1%    nan °C     2779 MHz
CPU1:	  7.1%    nan °C     2590 MHz
CPU2:	 14.1%    nan °C     2691 MHz
CPU3:	  2.1%    nan °C     2728 MHz
CPU4:	 10.1%    nan °C     3048 MHz
CPU5:	  8.2%    nan °C     3377 MHz
CPU6:	 16.2%    nan °C     2351 MHz
CPU7:	  9.9%    nan °C     3071 MHz
CPU8:	 10.9%    nan °C     3340 MHz
CPU9:	  4.1%    nan °C     3203 MHz
CPU10:	  9.3%    nan °C     3148 MHz
CPU11:	 12.6%    nan °C     3157 MHz
CPU12:	  8.3%    nan °C     3173 MHz
CPU13:	  4.0%    nan °C     2832 MHz
CPU14:	 10.3%    nan °C     3295 MHz
CPU15:	  4.2%    nan °C     3193 MHz

fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Snap package: no

Python: 3.9.1
psutil package: 5.7.3
platform package: 1.0.8
[...]

PR #590, opened 2023-10-30
Add zenpower for AMD CPU temps

fredinj:
#145 

So I couldn't get CPU temps of my AMD CPU with zenpower installed. This seems to fix it like it's mentioned in the issue.


AdnanHodzic:

PR #599, opened 2023-11-12
Create auto-cpufreq-sysvinit

AmerXz:
A sysvinit daemon for auto-cpufreq generated THANKS TO :- 

https://github.com/yunginnanet/sysvinit-service-generator fork of https://github.com/wyhasany/sysvinit-service-generator

It should be inside /etc/init.d/

### DIFF
--- a/scripts/auto-cpufreq-sysvinit
+++ b/scripts/auto-cpufreq-sysvinit
@@ -1,0 +1,98 @@
+#!/bin/sh
+### BEGIN INIT INFO
+# Provides:          auto-cpufreq
+# Required-Start:    $local_fs $network $named $time $syslog
+# Required-Stop:     $local_fs $network $named $time $syslog
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Description:       Automatic\ CPU\ speed\ \&\ power\ optimizer\ for\ Linux
+### END INIT INFO
+
+SCRIPT="/home/Abdulhafez/.scripts/Auto-cpufreq.sh"
+RUNAS=root
+
+PIDFILE=/var/run/auto-cpufreq.pid
+LOGFILE=/var/log/auto-cpufreq.log
+
+start() {
+  if [ -f $PIDFILE ] && [ -s $PIDFILE ] && kill -0 $(cat $PIDFILE); then
+    echo 'Service already running' >&2
+    return 1
+  fi
+  echo 'Starting service…' >&2
+  local CMD="$SCRIPT &> \"$LOGFILE\" & echo \$!"
+  su -c "$CMD" $RUNAS > "$PIDFILE"
+ # Try with this command line instead of above if not workable
+ # su -s /bin/sh $RUNAS -c "$CMD" > "$PIDFILE"
+ 
+  sleep 2
+  PID=$(cat $PIDFILE)
+    if pgrep -u $RUNAS -f $NAME > /dev/null
+    then
+      echo "$NAME is now running, the PID is $PID"
+    else
+      echo ''
+      echo "Error! Could not start $NAME!"
+    fi
+}
+
+stop() {
+  if [ ! -f "$PIDFILE" ] || ! kill -0 $(cat "$PIDFILE"); then
+    echo 'Service not running' >&2
+    return 1
+  fi
+  echo 'Stopping service…' >&2
+  kill -15 $(cat "$PIDFILE") && rm -f "$PIDFILE"
+  echo 'Service stopped' >&2
+}
+
+uninstall() {
+  echo -n "Are you really sure you want to uninstall this service? That cannot be undone. [yes|No] "
+  local SURE
+  read SURE
+  if [ "$SURE" = "yes" ]; then
+    stop
+    rm -f "$PIDFILE"
+    echo "Notice: log file was not removed: $LOGFILE" >&2
+    update-rc.d -f $NAME remove
+    rm -fv "$0"
+  else
+    echo "Abort!"
+  fi
+}
+
+status() {
+    printf "%-50s" "Checking auto-cpufreq..."
+    if [ -f $PIDFILE ] && [ -s $PIDFILE ]; then
+        PID=$(cat $PIDFILE)
+            if [ -z "$(ps axf | grep ${PID} | grep -v grep)" ]; then
+                printf "%s\n" "The process appears to be dead but pidfile still exists"
+            else    
+                echo "Running, the PID is $PID"
+            fi
+    else
+        printf "%s\n" "Service not running"
+    fi
+}
+
+
+case "$1" in
+  start)
+    start
+    ;;
+  stop)
+    stop
+    ;;
+  status)
+    status
+    ;;
+  uninstall)
+    uninstall
+    ;;
+  restart)
+    stop
+    start
+    ;;
+  *)
+    echo "Usage: $0 {start|stop|status|restart|uninstall}"
+esac


AmerXz (commented):
Run script For auto-cpufreq That can make sysvinit recognise auto-cpufreq 

Note You can Also use the runscript inside  auto-cpufreq/scripts/auto-cpufreq-s6 By @AnasR7

scripts/auto-cpufreq-run
AmerXz, line 1: Run script For auto-cpufreq That can make sysvinit recognise auto-cpufreq 

Note You can Also use the runscript inside  auto-cpufreq/scripts/auto-cpufreq-s6 By @AnasR7





/run

AdnanHodzic:
LGTM, but the part which I'm missing is how will the [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto-cpufreq-installer) detect if system user is running on is running on systemv and it's not refering to [install script anywhre](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/scripts/auto-cpufreq-install.sh)?

Or was there other you were envisioning of working?

scripts/auto-cpufreq-run
```diff
@@ -0,0 +1,3 @@
+!#/bin/bash
```
shadeyg56, line 1: remove this file since we already have one like it

shadeyg56 (changes requested):
Thanks for this pr.

I do agree we should remove the `auto-cpufreq-run`  since we already have the runfile in` auto-cpufreq/scripts/auto-cpufreq-s6` to avoid redundancy and bloat

Also to repeat what Adnan suggested, the [daemon installer script](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/scripts/auto-cpufreq-install.sh#L71C25-L71C25) should be updated to include this new sysvinit service. The problem I see is what we already support openrc which uses sysvinit by default, so they both use the `init` process name for PID 1, so we need a way to distinguish the difference. 

Something like this could work:
```sh
# snippet
elif [ "$(ps h -o comm 1)" = "init" ];then
    # Check if /etc/inittab file exists
    if [ -e "/etc/inittab" ]; then
       # sysvinit command here
    else
        # openrc stuff moved here
    fi
```

Additionally, the same change should be made to the [daemon removal script](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/scripts/auto-cpufreq-remove.sh#L55). If that doesn't make sense, please let me know and also feel free to comment if you have any alternative ideas since I am not as familiar with openrc or sysvinit :)

AmerXz:
It's deleted

scripts/auto-cpufreq-run
AmerXz: It's deleted

AdnanHodzic:
> It's deleted

:+1:, but this part is still missing:

> Also to repeat what Adnan suggested, the [daemon installer script](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/scripts/auto-cpufreq-install.sh#L71C25-L71C25) should be updated to include this new sysvinit service. The problem I see is what we already support openrc which uses sysvinit by default, so they both use the `init` process name for PID 1, so we need a way to distinguish the difference.
> 
> Something like this could work:
> 
> ```shell
> # snippet
> elif [ "$(ps h -o comm 1)" = "init" ];then
>     # Check if /etc/inittab file exists
>     if [ -e "/etc/inittab" ]; then
>        # sysvinit command here
>     else
>         # openrc stuff moved here
>     fi
> ```
> 
> Additionally, the same change should be made to the [daemon removal script](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/scripts/auto-cpufreq-remove.sh#L55). If that doesn't make sense, please let me know and also feel free to comment if you have any alternative ideas since I am not as familiar with openrc or sysvinit :)

For reference see how in [daemon installer script](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/scripts/auto-cpufreq-install.sh#L71C25-L71C25) it was done for `# Install script for openrc` or `# Install script for systemd` and refer to @shadeyg56 comment above.

AmerXz:
Sorry if I was late I want to take some time until I finish some few important things to do these days  and I need to organize myself a little bit.

I will test my fork repository and will give you what happens then .

 also thanks for these suggestions.

AdnanHodzic:
No worries and take your time

AmerXz:
```
# Install script for openrc / sysvinit
elif [ "$(ps h -o comm 1)" = "init" ];then
   mkdir /etc/inittab
    if [ -e "/etc/inittab" ]; then 
        echo -e "\n* Deploy auto-cpufreq sysvinit unit file"
        cp /usr/local/share/auto-cpufreq/scripts/auto-cpufreq-openrc /etc/init.d/auto-cpufreq
	
        echo -e "\n* "Reloading sysvinit manager configuration"
	service auto-cpufreq reload
 
        echo -e "\n* "Starting auto-cpufreq daemon (sysvinit) service"
        service auto-cpufreq start
        
        echo -e "\n* "Enabling auto-cpufreq daemon (sysvinit) service at boot"
	chkconfig auto-cpufreq-sysvinit on
      else
        echo -e "\n* Deploy auto-cpufreq openrc unit file"
        cp /usr/local/share/auto-cpufreq/scripts/auto-cpufreq-openrc /etc/init.d/auto-cpufreq
	echo -e "\n* "Starting auto-cpufreq daemon (openrc) service"
	rc-service auto-cpufreq start

	echo -e "\n* Enabling auto-cpufreq daemon (openrc) service at boot"
	rc-update add auto-cpufreqlocal/share/auto-cpufreq/scripts/auto-cpufreq-openrc /etc/init.d/auto-cpufreq
```


I Edited openrc script to be like that and I used this fedora resource in this link that helped me find sysvinit commands https://fedoraproject.org/wiki/SysVinit_to_Systemd_Cheatsheet

For now I didn't test the code 

tell me for any suggestions if needed

I'll see you guys next time

shadeyg56:
I will test in a VM with both init systems

phrygian666:
I am testing this right now on MX linux 23 Fluxbox edition (non systemd) and would like to post my observations from discord (sry for the ad-hoc writing style):

Mh, it seems to work, installation went good, when i started in monitor mode it reminded me to de-install tlp, I did that restarted, now I'm running in live mode, the temperatures are the same with a different monitoring tool, unplugging the power supply gets detected, the setting to use governor is the same like in cat /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor   from another terminal. The testing laptop is fairly old laptop: Not setting EPP (Not supported by system)  Warning: CPU turbo is not available seem plausible

I see I have to check if I get it installed permanently

There seems to be a problem in line 84 at the auto-cpufreq-install script

echo -e "\n* "Enabling auto-cpufreq daemon (sysvinit) service at boot"
    chkconfig auto-cpufreq-sysvinit on
    
 when running sudo auto-cpufreq -- install

   * Turn off bluetooth on boot


    * Deplo auto-cpufreq*remove script

There was a problem, couldn't determine GNOME Power Profiles Daemon 

-----------------Running auto-cpufreq daemon install script -------------------------
/usr/local/bin/auto-cpufreq-install: line 84: syntax error near unexpected token >> ( <<
/usr/local/bin/auto-cpufreq-install: line 84: echo -e "\n* "Enabling auto-cpufreq daemon (sysvinit) service at boot" '

--------------------- auto-cpufreq daemon installed and runnin--------------------------

....
I wrote that by hand from another pc


Thank you for your work on non-systemd compatibility . :)

phrygian666:
Further investing there seems to be at least some trouble with the the additional  (")  after "\n 
from line 78 - 89  

        echo -e "\n* "Reloading sysvinit manager configuration"
	service auto-cpufreq reload
 
        echo -e "\n* "Starting auto-cpufreq daemon (sysvinit) service"
        service auto-cpufreq start

        echo -e "\n* "Enabling auto-cpufreq daemon (sysvinit) service at boot"
	chkconfig auto-cpufreq-sysvinit on

and 

          echo -e "\n* "Starting auto-cpufreq daemon (openrc) service"

phrygian666:
After deleting the additional " the terminal output says it doesn't know chkconfig which doesn't seem to be used by mx linux fluxbox (not installed).

AmerXz:
Ok it seems that chkconfig command is deprecated years ago and we should use insteasd sysv-rc-conf it can be found in debian repos https://packages.debian.org/bookworm/sysv-rc-conf
 
source https://askubuntu.com/questions/221293/why-is-chkconfig-no-longer-available-in-ubuntu/277732#277732

phrygian666:
Alrighty, there are some errors I want to post here besides chkconfig, you might have come across them in your virtual machine:

when running auto-cpufreq --install

I also get these three errors:

mkdir: cannot create folder : "/etc/inittab": file exists already

* Reloading sysvinit manager configuration
auto-cpufreq: unrecognized service

* starting auto-cpufreq daemon (ssvinit) service
auto-cpufreq: unrecognized service

Again, thanks for your work :)

scripts/auto-cpufreq-install.sh
```diff
@@ -70,29 +70,25 @@ elif [ "$(ps h -o comm 1)" = "systemd" ];then
...
 elif [ "$(ps h -o comm 1)" = "init" ];then
-   mkdir /etc/inittab
+   ls /etc/inittab
     if [ -e "/etc/inittab" ]; then 
         echo -e "\n* Deploy auto-cpufreq sysvinit unit file"
-        cp /usr/local/share/auto-cpufreq/scripts/auto-cpufreq-openrc /etc/init.d/auto-cpufreq
-	
-        echo -e "\n* "Reloading sysvinit manager configuration"
-	service auto-cpufreq reload
+        cp /usr/local/share/auto-cpufreq/scripts/auto-cpufreq-sysvinit /etc/init.d/auto-cpufreq
```
nixon87, line 77: I think you need a 
`chmod +x /etc/init.d/auto-cpufreq`
here
The copying to /etc/init.d works on my system but the script is not executable.

nixon87:
Just for a heads up: as written in the review  the copying when running auto-cpufreq --install seems to work now. When adding chmod +x /etc/init.d/auto-cpufreq and restart the os the shell script runs as a system service at reboot. But when running auto-cpufreq --stats the daemon doesn't seem to be active. I think it's close to working but sadly I am no expert at all. I guess the sysvinit script needs some revisit and maybe the things pointed out by [shadeyg56](https://github.com/shadeyg56) regarding the daemon removal script. Maybe Shadeyg56 can take a look at the state right now?

AdnanHodzic:
@shadeyg56 could you please take a final look and confirm if it's all working, if it's let's merge this. Thanks!

nixon87:
As written above I think it's not quite there yet, at least on my machine the auto-cpufreq daemon isn't active. The init system script gets started however (after the suggested changes).

shadeyg56:
Haven't been able to test in a VM since `sudo auto-cpufreq --install` does not work in VMs so the OS would need to be installed natively. @AmerXz have you tested the changes?

AmerXz (commented):
line 92 has extra "local/share/auto-cpufreq/scripts/auto-cpufreq-openrc /etc/init.d/auto-cpufreq" added there by mistake 

This is fixed in the more updated commit

scripts/auto-cpufreq-install.sh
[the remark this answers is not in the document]
AdnanHodzic: I suggest you prepend "WIP"  (work in progress) to title of this PR, and then please remove WIP part & announce it when it's ready for review.

Thanks!

AmerXz:
I did some test to the code and it ran perfectly fine on mx-linux with sysvinit even after reboot , But  I couldn't test it in openrc distro .

I tried to install artix with openrc but the installation was broken so I need another distro or maybe other iso version .

rakshasha-medhi:
> I did some test to the code and it ran perfectly fine on mx-linux with sysvinit even after reboot , But I couldn't test it in openrc distro .
> 
> I tried to install artix with openrc but the installation was broken so I need another distro or maybe other iso version .

Hi, I'm looking to use auto-cpufreq with MX-Linux as well but I still got the same error message in your earlier posts:
*System has not been booted with systemd as init system (PID 1). Can't operate.

how did you manage to get it work with MX-Linux and sysvinit system? would love to learn more to get this working.
or is this still undergoing tests? I notice the status is WIP.
Thanks for the great tool !

AmerXz:
Did you clone from `patch-2` branch in my fork repo ?

AdnanHodzic:
IIRC @Angel-Karasu recently made some changes to make auto-cpufreq work with MX Linux?

Regardless, @AmerXz if this is anywhere close to merging (if it works) maybe it's good idea to start the review/merge process?

AmerXz:
It works on mx-linux but I didn't test it on openrc distro

PR #606, opened 2023-11-22
update README to match changes to the AUR package

MusicalArtist12:
Note the inclusion of my username - i'm unsure of this specific aspect. Also take note of the request to leave an issue on this repo - the only other method for me to know if something is wrong is through AUR comments. 

I moved the Gnome Power Profiles Daemon issue from `#Troubleshooting` to the `###AUR package` as this made more sense to me. it is also baked into the arch installer. 

I also made note of when both AUR repos where last updated- I don't have control over the git repo so its worth mentioning the difference. FWIW its also marked as out of date on the AUR and was last updated last year.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,12 +122,20 @@ sudo snap install auto-cpufreq
 
 ### AUR package (Arch/Manjaro Linux)
 
-*The AUR packages below are often unmaintained & have issues*! Unless you see evidence of good recent maintenance, use the [auto-cpufreq-installer](#auto-cpufreq-installer) instead as otherwise you'll run into errors (e.g., [#471](https://github.com/AdnanHodzic/auto-cpufreq/issues/471)). If you still choose to install via AUR, see the [Troubleshooting](#aur) section for solved known issues.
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
> Note the inclusion of my username

I think this is good, even preferable as if anyone has any questions they can reach out directly to you, either in Discussion/Issues section or Discord server. 

> I moved the Gnome Power Profiles Daemon issue from #Troubleshooting to the ###AUR package as this made more sense to me. it is also baked into the arch installer.

Ack, just be aware this will also be mentioned here:

> 1: power_helper.py script (Snap package install only)
When installing auto-cpufreq via [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/blob/9c5701eeca3cc7986058241fb16ece3ed1187fa2/README.md#auto-cpufreq-installer), if it detects the [GNOME Power Profiles service](https://twitter.com/fooctrl/status/1467469508373884933) is running, it will automatically disable it. Otherwise, that daemon will cause conflicts and various other performance issues.

> I also made note of when both AUR repos where last updated- I don't have control over the git repo so its worth mentioning the difference. FWIW its also marked as out of date on the AUR and was last updated last year.

Ack, LGTM! 

Thank you for your contribution, I'll update the [v2.1.0 release ](https://github.com/AdnanHodzic/auto-cpufreq/releases) and add you there, instead of making a new v2.1.1.

PR #609, opened 2023-11-25
nix: bump version to 2.1.0 and update lock file

shadeyg56:
Simple enough, just updating the version in the nix package to the newest and went ahead and updated the lock file while I was at it

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


AdnanHodzic:
It would be great if version could be picked up as it is for [snap package](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/snap/snapcraft.yaml#L34), this way we only need to update one file and it's propagated for all packages ...

PR #619, opened 2023-12-17
Add support for configuring/overriding "Energy Performance Preference (EPP)"

murbot:
These changes add support for configuring the energy performance preference value to be used when on charger and battery, via the auto-cpufreq config file. I've updated the readme and example config with examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ By default, auto-cpufreq does not use the config file! If you wish to use it, th
 # preferred governor
 governor = performance
 
-# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
+# EPP: see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
 energy_performance_preference = performance
 
 # minimum cpu frequency (in kHz)
@@ -310,7 +310,7 @@ turbo = auto
 # preferred governor
 governor = powersave
 
-# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
+# EPP: see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
 energy_performance_preference = power
 
 # minimum cpu frequency (in kHz)
--- a/README.md
+++ b/README.md
@@ -286,6 +286,9 @@ By default, auto-cpufreq does not use the config file! If you wish to use it, th
 # preferred governor
 governor = performance
 
+# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
+energy_performance_preference = performance
+
 # minimum cpu frequency (in kHz)
 # example: for 800 MHz = 800000 kHz --> scaling_min_freq = 800000
 # see conversion info: https://www.rapidtables.com/convert/frequency/mhz-to-hz.html
@@ -306,6 +309,9 @@ turbo = auto
 # see available governors by running: cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors
 # preferred governor
 governor = powersave
+
+# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
+energy_performance_preference = power
 
 # minimum cpu frequency (in kHz)
 # example: for 800 MHz = 800000 kHz --> scaling_min_freq = 800000
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ By default, auto-cpufreq does not use the config file! If you wish to use it, th
 # preferred governor
 governor = performance
 
-# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
+# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
 energy_performance_preference = performance
 
 # minimum cpu frequency (in kHz)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -286,6 +286,9 @@ By default, auto-cpufreq does not use the config file! If you wish to use it, th
 # preferred governor
 governor = performance
 
+# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
```
AdnanHodzic, line 289: This is incorrect, if you want to see available preferences you should do:
`cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences`

Please update for every other comment where:

```
# see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
```

is mentioned.

auto_cpufreq/core.py
```diff
@@ -676,8 +676,13 @@ def set_powersave():
...
         and Path("/sys/devices/system/cpu/intel_pstate/hwp_dynamic_boost").exists() is False
     ):
-        run("cpufreqctl.auto-cpufreq --epp --set=balance_power", shell=True)
-        print('Setting to use: "balance_power" EPP')
+        if conf.has_option("battery", "energy_performance_preference"):
+            epp = conf["battery"]["energy_performance_preference"]
+            run(f"cpufreqctl.auto-cpufreq --epp --set={epp}", shell=True)
+            print(f'Setting to use: "{epp}" EPP')
+        else:
+            run("cpufreqctl.auto-cpufreq --epp --set=balance_power", shell=True)
```
AdnanHodzic, line 684: Do we now get in situation where if nothing was defined for `energy_performance_preference` in config file (or user isn't using config file), `balance_performance` will be used in both cases when computer is in charging & on power/battery? Or am I missing something?

README.md
murbot: Sorry, silly mistake. I've corrected this.

auto_cpufreq/core.py
murbot: I don't think so, unless I'm missing something. The _if_ block only executes if the configuration option is present. Otherwise, the _else_ block executes and the value is set to `balance_power`.

I've been testing on my machine and it works well.

AdnanHodzic: Never-mind I overlooked what was happened here.

AdnanHodzic:
I love what's happening here, and instead of having it as "improvement" I want to have this as one of the features for next release. I could already merge it as it is, but I have another small request/suggestion which would make it perfect. 

Could you as part of this PR, make it visible as part of output what value is used, i.e `Energy Performance Preference (EPP) in use: balance_performance`. It could be added under `CPU frequency scaling` section, ideally under i.e: `Average temp. of all cores` field. Think it would be of great value to have this.

What do you think, would this be possible from your side?

AdnanHodzic:
> Setting to use: "power" EPP
> 
> Do you have another change in mind?

Yes, this line could be re-used for this purpose. Reason why I mentioned it is because I used to see this line. Now with latest change I don't see anything being displayed as part of it.

This is what `--stats` looks like with change on your branch, even after using config file (same thing in GUI of course) and explicitly defining EEP override.  

```
------------------------------------------------------------------------------

Linux distro: Ubuntu 23.10 Mantic Minotaur
Linux kernel: 6.5.0-14-generic
Cores: 8
Architecture: x86_64
Driver: intel_pstate

CPU max frequency: 1800 MHz
CPU min frequency: 400 MHz

CPU0      4.0%        35 °C      1800 MHz
CPU1      3.0%        36 °C      1800 MHz
CPU2      2.0%        34 °C       400 MHz
CPU3     11.2%        36 °C      1800 MHz
CPU4      6.9%        35 °C      1800 MHz
CPU5     25.7%        36 °C      1800 MHz
CPU6      5.9%        34 °C      1800 MHz
CPU7      7.1%        36 °C      1801 MHz

CPU fan speed: 0 RPM

Battery is: charging

Total CPU usage: 2.9 %
Total system load: 0.92
Average temp. of all cores: 35.25 °C 

High CPU load (load average: 0.92, 1.86, 2.08)
setting turbo boost: on

```

Also on this topic, let's make small changes to config file, so in front of `# EPP: see available preferences by running` lines it says `# EPP: see available preferences by running`

AdnanHodzic:
> I pushed a new commit adding the "EPP: " to the config file comments.

:+1: 

> Note, that this condition was in the code prior to my additions.

I'm aware, it's just now that I realized now I'm no longer seeing it.

>  I assume you're not seeing the EPP output because the following condition evaluates to false. 
> In short, it is not shown because you're on an Intel processor, whereas I'm on AMD.

    if (
        Path("/sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference").exists()
        and Path("/sys/devices/system/cpu/intel_pstate/hwp_dynamic_boost").exists() is False
    ):

Indeed `/sys/devices/system/cpu/intel_pstate/hwp_dynamic_boost` will return 0 in my case.

Then let's do this, since by default "EPP" line will come as enabled (not commented) in config file. Where in this case, EPP is not showing up, let's extend the code to something like:

```
if (
    os.path.exists("/sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference")
    and not os.path.exists("/sys/devices/system/cpu/intel_pstate/hwp_dynamic_boost")
):
    value = int(
        os.popen("cat /sys/devices/system/cpu/intel_pstate/hwp_dynamic_boost").read()
    )
    if value == 0:
        print("(EPP) in use: dynamic boost is disabled")
    else:
        print("(EPP) in use: {epp}")
else:
    print("(EPP) in use: Dynamic boost is not supported")
``` 
    
Feel free to amend code block, it's just the first thing I could think of.

AdnanHodzic:

Issue #563, opened 2023-09-27
Spamming of 'line 105: echo: write error: Device or resource busy' in journal


Fxzzi:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? **Yes**
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)? **No**
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)? **Yes**
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)? **Yes**

### Error output:
```text
line 105: echo: write error: Device or resource busy
```
---

### System information:

Add/paste output of:

```
-------------------------------------------------------------------------------

Linux distro: Arch Linux  
Linux kernel: 6.5.5-arch1-1
Processor: AMD Ryzen 5 PRO 7530U with Radeon Graphics
Cores: 12
Architecture: x86_64
Driver: amd-pstate-epp

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 4546 MHz
CPU min frequency: 400 MHz

Core	Usage	Temperature	Frequency
CPU0      0.0%        59 °C       400 MHz
CPU1      2.9%        59 °C      2665 MHz
CPU2      4.0%        59 °C      2298 MHz
CPU3      0.0%        59 °C       400 MHz
CPU4      2.0%        59 °C       400 MHz
CPU5      0.0%        59 °C      2507 MHz
CPU6      1.0%        59 °C      2327 MHz
CPU7      0.0%        59 °C      1916 MHz
CPU8      0.0%        59 °C       400 MHz
CPU9      1.0%        59 °C      2762 MHz
CPU10      1.0%        59 °C      3576 MHz
CPU11      1.0%        59 °C      2096 MHz

CPU fan speed: 0 RPM

auto-cpufreq version: 1.9.9 (git: fdb20f5e)

Python: 3.11.5
psutil package: 5.9.5
platform package: 1.0.8
click package: 8.1.7
distro package: 1.8.0

Computer type: Notebook
Battery is: charging

[...]

PR #627, opened 2024-01-20
Ignore "Device or resource busy" error message when writing value

hingen:
Potential fix for the issue described in #563 by ignoring "write error: Device or resource busy" error message when writing a value.

### DIFF
--- a/scripts/cpufreqctl.sh
+++ b/scripts/cpufreqctl.sh
@@ -105,8 +105,8 @@ function write_value () {
     ERROR_MESSAGE=$((echo $VALUE > $FLNM) 2>& 1)
     EXIT_STATUS=$?
 
-    if ! [[ -z $ERROR_MESSAGE || $ERROR_MESSAGE == *"write error: Device or resource busy."* ]]; then
-      echo ERROR_MESSAGE >& 2
+    if ! [[ -z $ERROR_MESSAGE || $ERROR_MESSAGE == *"write error: Device or resource busy"* ]]; then
+      echo $ERROR_MESSAGE >& 2
     fi
 
     return $EXIT_STATUS
--- a/scripts/cpufreqctl.sh
+++ b/scripts/cpufreqctl.sh
@@ -103,10 +103,13 @@ function driver () {
 function write_value () {
   if [ -w $FLNM ]; then
     ERROR_MESSAGE=$((echo $VALUE > $FLNM) 2>& 1)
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Thank you for the PR, since these changes seem strictly AMD oriented, I've asked [on auto-cpufreq discord channel](https://discord.gg/Sjauxtj6kH) if someone could test these changes.

Tresillo2017:
Tried on my laptop with a Ryzen 7 and looks like it's not apearing anymore on journalctl but I have the same problem where on auto-cpufreq it shows that it can set EPP because it not supported
image

AdnanHodzic:
@Tresillo2017 this issue was fixed with #631 can you make sure you have pulled the latest changes on `master` branch?

shadeyg56:
I'm not sure this is related to #631 since that was a fix for Intel CPUs.

You're using the `acpi-cpufreq` driver which does not support EPP, so this is not a bug. If you would like EPP support, you'll need to switch to using `amd-pstate-epp`

Tresillo2017:
> I'm not sure this is related to #631 since that was a fix for Intel CPUs.
> 
> You're using the `acpi-cpufreq` driver which does not support EPP, so this is not a bug. If you would like EPP support, you'll need to switch to using `amd-pstate-epp`

And how can I do that? By not typing anything into the grub parameters?

AdnanHodzic:
> > I'm not sure this is related to #631 since that was a fix for Intel CPUs.
> > You're using the `acpi-cpufreq` driver which does not support EPP, so this is not a bug. If you would like EPP support, you'll need to switch to using `amd-pstate-epp`
> 
> And how can I do that? By not typing anything into the grub parameters?

Explained as part of: https://github.com/AdnanHodzic/auto-cpufreq/#troubleshooting

Tresillo2017:
> > > I'm not sure this is related to #631 since that was a fix for Intel CPUs.
> > > You're using the `acpi-cpufreq` driver which does not support EPP, so this is not a bug. If you would like EPP support, you'll need to switch to using `amd-pstate-epp`
> > 
> > And how can I do that? By not typing anything into the grub parameters?
> 
> Explained as part of: https://github.com/AdnanHodzic/auto-cpufreq/#troubleshooting

That was a really quick reply, thx for that. And now it's working as supposed, let's se if my battery life improves I was getting 3-4h of battery life out of the box on a thinkpad e16 amd.

Issue #628, opened 2024-01-21
GUI not opening if not connected to internet


hingen:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted?  Yes
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)? NA
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)? Yes
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)? Yes

### Error output:
No output, CLI or GUI.

---

### System information:

```
-------------------------------------------------------------------------------

Linux distro: Arch Linux rolling n/a
Linux kernel: 6.7.0-arch3-1
Processor: AMD Ryzen 7 5800H with Radeon Graphics
Cores: 16
Architecture: x86_64
Driver: amd-pstate-epp

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 4463 MHz
CPU min frequency: 400 MHz

Core    Usage   Temperature     Frequency
CPU0      5.7%        71 °C      3315 MHz
CPU1      0.0%        71 °C       400 MHz
CPU2      2.0%        71 °C      3300 MHz
CPU3      1.0%        71 °C      4322 MHz
CPU4      1.0%        71 °C       400 MHz
CPU5      1.0%        71 °C      3166 MHz
CPU6      1.0%        71 °C      3361 MHz
CPU7      0.0%        71 °C       400 MHz
CPU8      0.0%        71 °C       400 MHz
CPU9      0.0%        71 °C       400 MHz
CPU10      4.0%        71 °C       400 MHz
CPU11      0.0%        71 °C       400 MHz
CPU12      1.0%        71 °C      3679 MHz
CPU13      0.0%        71 °C       400 MHz
CPU14     12.7%        71 °C      3369 MHz
CPU15      0.0%        71 °C      3257 MHz

auto-cpufreq version: 2.1.0 (git: fe29620)

Python: 3.11.6
psutil package: 5.9.8
platform package: 1.0.8
click package: 8.1.7
distro package: 1.9.0

[...]

PR #629, opened 2024-01-21
Change update checking to return "no update" instead of exiting process on server connection error

hingen:
Issue described in  #628


AdnanHodzic:

PR #631, opened 2024-01-23
Fix dynamic boosting being incorrectly diagnosed

PhoenixCausesOof:
Previously, `auto-cpufreq` would only check if the `intel_pstate/hwp_dynamic_boost` file existed to determine if EPP was supported. This didn't account for the fact that the file's contents could be `0`, so it incorrectly assumed dynamic boosting was enabled.

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -901,7 +901,7 @@ def set_performance():
     )
 
 
-     if Path("/sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference").exists() is false:
+    if Path("/sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference").exists() is false:
         print('Not setting EPP (not supported by system)')
     else:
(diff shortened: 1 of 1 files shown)


PhoenixCausesOof:
Ah, my bad. Never really programmed Python before, so I felt I might it a little try.

Anyway, the problem (thankfully) was just an extra whitespace character (thanks, Python). I removed it, so should be working now.

PhoenixCausesOof:
There we go. It was my first time programming Python, ever, so forgive for that.

Anyway, tested on my machine as well (didn't know that was possible, was a GUI user) and it works now.

By the way, the two issues I faced were:
1. Boolean literals are `True` and `False`, not `true` and `false`.
2. Indentation (more specifically, BECAUSE OF A SINGLE WHITESPACE CHARACTER, my code refused to compile).

AdnanHodzic:
LGTM! It's to good to have EPP showing up again on Intel hardware, thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #637, opened 2024-02-03
add battery charge thresholds

PurpleWazard:
hello, ive been using auto-cpufreq since i moved from TLP however i miss being able to set charge thresholds to prolong battery health. i would like to add this feature to auto-cpufreq, this allows you to set charge thresholds to support laptops. currently this only supports lenovo laptops and thinkpads however other laptops can be added easily. this feature is by default turned off users who want to use it will have to specify.

to change and enable the battery thresholds add these lines to /etc/auto-cpufreq.conf under [battery]

enable_thresholds = true
start_threshold = 0 
stop_threshold = 100

default thresholds are 0 and 100

auto-cpufreq --daemon will set the charge thresholds on boot
auto-cpufreq --monitor will set and print the charge thresholds once 

there was an issue with --live so i remove it from live. 

this feture requires the correct kernel modules to be loaded to identify the laptop type. example for thinkpads the thinkpad_acpi must be listed in `lsmod | grep thinkpad_acpi`  to add the thinkpad module to thinkpads run `modprobe thinkpad_acpi`

there may be some bugs however i fixed all the ones i could find. 

this is also my first pull request, suggestions are appreciated.

### DIFF
--- a/auto-cpufreq.conf-example
+++ b/auto-cpufreq.conf-example
@@ -45,3 +45,17 @@ energy_performance_preference = power
 
 # turbo boost setting. possible values: always, auto, never
 turbo = auto
+
+# experimental 
+
+# Add battery charging threshold (currently only available to Lenovo)
+# checkout README.md for more info
+
+# enable thresholds true or false
+#enable_thresholds = true
+#
+# start threshold (defaults to 0 ) can be 0 - 100
+#start_threshold = 0
+#
+# stop threshold (defaults to 100) this value must be greater or equal to 65
+#stop_threshold = 100
--- a/auto_cpufreq/battery_scripts/battery.py
+++ b/auto_cpufreq/battery_scripts/battery.py
@@ -18,7 +18,6 @@ def battery_start_threshold():
     conf = get_config()
     if conf.has_option("battery", "start_threshold"):
         start_threshold = conf["battery"]["start_threshold"]
-        print(f"start {type(start_threshold)} {start_threshold}")
         return int(start_threshold)
     else:
         return 0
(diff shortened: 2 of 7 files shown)


PurpleWazard:
i found a bug with reading values from the conf file im currently working on it

AdnanHodzic:
From code POV, it LGTM. 

I have 2 small requests:


**1:** 

Only thing I would ask for is to update `auto-cpufreq.conf-example` file with commented examples, i.e:

```
# Add battery charging threshold (currently only available to Lenovo)
#enable_thresholds = true
#start_threshold = 0
#stop_threshold = 100
```

and update README file accordingly, maybe also add this part, just rephrase it if needed:

> this feture requires the correct kernel modules to be loaded to identify the laptop type. example for thinkpads the >thinkpad_acpi must be listed in lsmod | grep thinkpad_acpi to add the thinkpad module to thinkpads run modprobe >thinkpad_acpi

**2:**

Code wise, it would be good if `stop_threshold = 80` following change was made, it would stop charging at 80%, as currently if 80 is set, it'll stop charging at 79%.

shadeyg56:
Yeah, LGTM as well. Of course, I can't test it since I do not have a Thinkpad laptop, but it seems to be working.

> and update README file accordingly,

This was going to be my main comment. We should include a new section in the README about battery thresholds including instructions, list of supported devices, and information on how to request that your device be added. I don't mind doing this myself since you might be as familiar with the structure of the README file.

PurpleWazard:
> Code wise, it would be good if `stop_threshold = 80` following change was made, it would stop charging at 80%, as currently if 80 is set, it'll stop charging at 79%.
from what im understanding is when you set stop threshold to 80 your battery says it stops at 79% instead?  this could be your windows manager or desktop envierment, ive noticed when reading battery presentage is could be + or - 1% 
or it could be that there is always a small load on the battery when its pluged in. example on my thinkpad with battery thresholds set while pluged in the battery precentage dropped ~7% over the course of  3 hours 
my suggestion to your problem is set the stop threshold to 81 

as for the readme and conf example sure i can add it and shadyg56 if you would like to help me out that would be great!

AdnanHodzic:
Clear, let's merge this and @shadeyg56 I hope you will create a follow up PR to update the README.

@PurpleWazard  thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #644, opened 2024-02-12
Update README with new battery thresholds feature 

shadeyg56:
Apologies for the mess in the last attempt I made lol. The rebase went horribly wrong.

Added a new section for the battery thresholds as discussed in #637

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ Example of `auto-cpufreq --stats` CLI output
     * [Update - auto-cpufreq update](#update---auto-cpufreq-update)
     * [Remove - auto-cpufreq daemon](#remove---auto-cpufreq-daemon)
     * [stats](#stats)
+* [Battery charging thresholds](#battery-charging-thresholds)
+  * [Supported Devices](#supported-devices)
+  * [Battery config](#battery-config)
 * [Troubleshooting](#troubleshooting)
     * [AUR](#aur)
 * [Discussion](#discussion)
@@ -93,6 +96,7 @@ Only devices with an Intel, AMD, or ARM CPU are supported. This tool was develop
   * CPU temperature in combination with CPU utilization/load (to prevent overheating)
   * System load
 * Automatic CPU & power optimization (temporary and persistent)
+* Settings battery charging thresholds (limited support)
 
 ## Installing auto-cpufreq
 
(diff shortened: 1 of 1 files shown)


PurpleWazard:
Some thing you will want to add is  the stop charge threshold can not be below 65 ( at least with thinkpad_acpi ) I've had errors having it lower than 65

shadeyg56:
>  the stop charge threshold can not be below 65 ( at least with thinkpad_acpi )

Hm, I can't find any documentation on this, so maybe this needs to be explored further?
Would you mind sharing the errors you get?

PurpleWazard:
I started building a test however I won't be able to finish it until tonight I'll reply back with results

PurpleWazard:
So it seems my test is working without issue so I'm going to do some investigating

AdnanHodzic:
Maybe it would also be a good idea to first merge #645 ?

shadeyg56:
Agreed.

README.md
```diff
@@ -465,6 +469,31 @@ If the daemon has been installed, live stats of CPU/system load monitoring and o
 
 `auto-cpufreq --stats`
 
+## Battery charging thresholds
+
+As of v2.2.0, battery charging thresholds can be set in the config file. This enforces your battery to start and stop charging at defined values
```
AdnanHodzic, line 474: Let's make turn [v2.0.0](https://github.com/AdnanHodzic/auto-cpufreq/releases) into a link.

AdnanHodzic: Would also be good to include  [this](https://github.com/AdnanHodzic/auto-cpufreq/pull/645#issuecomment-1942159310). Maybe as a last line for this section to say something like: _"Please note: only --monitor and --stats have battery threshold function"_

Or maybe it could be omitted, my only concern is that folks will ask it as a question. Alternatively it could be added as a comment for config file as part of battery threshold?

README.md
```diff
@@ -465,6 +469,31 @@ If the daemon has been installed, live stats of CPU/system load monitoring and o
...
+
+### Supported devices
+
+* **Lenovo ThinkPad** (thinkpad_acpi)*
+* **Lenovo IdeaPad** (ideapad_acpi)*
+
+***Please note, your laptop must have an installed ACPI kernel driver specific to the manufacturer.** To check if you have the correct module installed and loaded run `lsmod [module]`
+
+**To request that your device be supported, please open an [issue](https://github.com/AdnanHodzic/auto-cpufreq/issues/new). In your issue, make us aware of the driver that works with your laptop**
```
AdnanHodzic, line 483: Let's turn this into a link as well: [of the driver](https://github.com/AdnanHodzic/auto-cpufreq/#troubleshooting)

AdnanHodzic:
#645 merged, added a few comments, nothing major other then that LGTM!

PurpleWazard:
It all looks good to me @AdnanHodzic suggested that we clarify that --live doesn't call the battery functions. 
Is was because I had having issues when I ran --live with the battery functions however since I've rewritten the battery scripts I tested it with --live and I haven't had an issue.

I can add that change before or after the new release if needed

AdnanHodzic:
@PurpleWazard I would rather have it before and hence part of 2.2.0 release.

PurpleWazard:
It is done

AdnanHodzic:
Removed this comment because this is now addressed with merge of #646

>Would also be good to include https://github.com/AdnanHodzic/auto-cpufreq/pull/645#issuecomment-1942159310. Maybe as a last line for this section to say something like: "Please note: only --monitor and --stats have battery threshold function"
>
>Or maybe it could be omitted, my only concern is that folks will ask it as a question. Alternatively it could be added as a comment for config file as part of battery threshold?

@shadeyg56 then let's add those 2 links I suggested and it's ready to merge.

README.md
shadeyg56: I think there's some confusion here. The driver in this case, refers to the acpi driver that allows the kernel to interact with battery, bluetooth, wifi radio, etc... so like thinkpad_acpi and ideapad_acpi, not the CPU scaling driver

AdnanHodzic: Oh, right, then let's disregard this one.

README.md
AdnanHodzic: Since I got a moment now to do the release, I'll just merge the changes and make this update myself.

shadeyg56: Done. The link should work after the release is made, since it links directly to tag v2.2.0

AdnanHodzic:
LGTM, thank you for your contribution will make sure to make this PR as part of release notes for 2.2.0.

PR #659, opened 2024-02-27
Create workflows for Linux builds and Nix flake builds

shadeyg56:
This was briefly discussed in the Discord

2 new files are added in `.github/workflows`

`build-linux.yml` contains a workflow that installs Python and runs `sudo ./auto-cpufreq-installer` to make sure it installs correctly.

`build-nix.yml` contains a workflow that installs Nix and runs `nix build` to ensure that the Flake builds on every push/PR

Lastly, I added status badges to the top of the README file that show whether or not the builds are passing or failing. This is useful in-case a failed build makes it to master, as seen below
20240227_14h24m34s_grim

We could also probably add one for the Snap package, but I wasn't sure if this would be wanted due to the ambiguity over the support of the Snap package as of late

Let me know what you think!


AdnanHodzic:
LGTM! 

> We could also probably add one for the Snap package, but I wasn't sure if this would be wanted due to the ambiguity over the support of the Snap package as of late

Think this would be a great idea, especially since I wanted to do the same thing but never got to.

PR #663, opened 2024-03-07
Rework config and reload config on file change/creation/deletion

shadeyg56:
This PR reworks the config into its own class/file and introduces config reloads when the config file is changed, created, or deleted. This also introduces the utils folder to remove some clutter from `core.py`

The file watching is done using the `pyinotify` package which I've added as a Nix dependency and Poetry dependency. All file watching is done in a separate thread so that the daemon isn't blocked

Rather than using the old `get_config()` function which relied on setting a static attribute to the function, this uses a new `_Config()` class, but the config is actually accessed by importing `config` from the `config.py` file. This allows the config to be accessed from any file, basically as a singleton. `_Config()` should NEVER be accessed directly

This change can be tested by running the daemon, running `auto-cpufreq --stats`, then editing your config file. You should see the changes take place live in the stats window without any sort of daemon reload.

PR #672, opened 2024-04-02
Add load from user config in XDG_CONFIG_HOME if available

braun-steven:
This update introduces the flexibility to load the configuration file from multiple locations, prioritizing user preferences and system standards. Previously, the configuration was strictly read from a hardcoded system path (`/etc/auto-cpufreq.conf`). Now, the application first checks if the user has specified a configuration file path via command line arguments. If not, it looks for a configuration file in the user's config directory (`$XDG_CONFIG_HOME/auto-cpufreq/auto-cpufreq.conf`). If neither is found, it defaults to the original system-wide configuration file.

This allows users to add their auto-cpufreq configuration to their dotfiles.

In addition, since providing a config file path via `--config` is now an explicit choice, an error is raised if this path is not a valid file.

### Old behavior
```sh
auto-cpufreq --stats # Looks for config in "/etc/auto-cpufreq.conf"
auto-cpufreq --config <CONFIG> --stats # Looks for config in <CONFIG>
auto-cpufreq --config <INVALID_PATH> --stats  # No error/warning
```

### New behavior
```sh
auto-cpufreq --stats # Looks for config in (A) "$XDG_CONFIG_HOME/auto-cpufreq/auto-cpufreq.conf" then (B) "/etc/auto-cpufreq.conf"
auto-cpufreq --config <CONFIG> --stats # Looks for config in <CONFIG>
auto-cpufreq --config <INVALID_PATH> --stats  # Error
```

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -105,7 +105,7 @@ def find_config_file(args_config_file):
     system_config_file = "/etc/auto-cpufreq.conf"
 
     # If config file is provided as a command line argument but is invalid, exit with error message
-    if args_config_file != None and args_config_file is not "":
+    if args_config_file is not None:
         if not os.path.isfile(args_config_file):
             print(f"Config file specified with '--config {args_config_file}' not found.")
             sys.exit(1)
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -104,14 +104,14 @@ def find_config_file(args_config_file):
     user_config_file = os.path.join(user_config_dir, "auto-cpufreq/auto-cpufreq.conf")
     system_config_file = "/etc/auto-cpufreq.conf"
 
-    # If config file is provided as a command line argument but is invalid, exit with error message
-    if args_config_file is not None:
-        if not os.path.isfile(args_config_file):
+    if args_config_file is not None: # (1) Command line argument was specified
+        # Check if the config file path points to a valid file
+        if os.path.isfile(args_config_file):
+            return args_config_file
+        else:
+            # Not a valid file
             print(f"Config file specified with '--config {args_config_file}' not found.")
             sys.exit(1)
-
-    if args_config_file and os.path.isfile(args_config_file):  # (1) Command line argument
-        return args_config_file
     elif os.path.isfile(os.path.join(user_config_file)):  # (2) User config file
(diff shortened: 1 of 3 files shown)


auto_cpufreq/core.py
```diff
@@ -84,6 +84,40 @@ def file_stats():
...
+    user_config_file = os.path.join(user_config_dir, "auto-cpufreq/auto-cpufreq.conf")
+    system_config_file = "/etc/auto-cpufreq.conf"
+
+    # If config file is provided as a command line argument but is invalid, exit with error message
+    if args_config_file is not None:
+        if not os.path.isfile(args_config_file):
+            print(f"Config file specified with '--config {args_config_file}' not found.")
+            sys.exit(1)
+
+    if args_config_file and os.path.isfile(args_config_file):  # (1) Command line argument
```
shadeyg56, line 113: `os.path.isfile(args_config_file)` in line 113 is redundant since it would be caught by line 109

shadeyg56 (changes requested):
I left one small comment in the review. I think overall this is a good change, but the config implementation is currently being reworked in #663 so this PR is going to conflict with that. It might be best to merge this PR to a different branch and then I can handle the merge myself. 

I'll let @AdnanHodzic leave his thoughts.

AdnanHodzic:
> It might be best to merge this PR to a different branch and then I can handle the merge myself.

@shadeyg56, yes let's do this. As I said in PM, I'm out traveling for work for next 2 weeks and don't have my personal laptop with me to test/look into this properly. 

Hence please feel free to merge it if it makes sense to you.

auto_cpufreq/core.py
braun-steven: I've merged the two conditions: https://github.com/AdnanHodzic/auto-cpufreq/pull/672/commits/2b3e9d9574b82c6bfab4f8fb2bcd72646eb5a72b

shadeyg56 (approved):

shadeyg56:
Looks good! I'm merging this into the `config-conflict` branch and I will merge these changes into #663 so that all of it can be merged into `master` at the same time.

Thanks for your contribution. You'll be credited as part of the next [release](https://github.com/adnanhodzic/auto-cpufreq/releases)

Issue #667, opened 2024-03-30
No "Warning: CPU turbo is not available" when using amd-pstate-epp


LDprg:
### Description
There is a warning about CPU turbo is not available when using amd-pstate-epp (amd-pstate in active mode). This warning doesn't make sense, since amd-pstate-epp is managing the CPU turbo within the driver. Maybe there should be a message about CPU turbo managed by the driver, so people are not confused.

### Error output:
See System information.

### System information:

```
-------------------------------------------------------------------------------

Linux distro: NixOS 24.05 Uakari
Linux kernel: 6.8.1
Processor: AMD Ryzen 7 5800H with Radeon Graphics
Cores: 16
Architecture: x86_64
Driver: amd-pstate-epp

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 4463 MHz
CPU min frequency: 400 MHz

Core	Usage	Temperature	Frequency
CPU0      3.0%        56 °C      3165 MHz
CPU1      0.0%        56 °C       400 MHz
CPU2      1.0%        56 °C      3193 MHz
CPU3      0.0%        56 °C       400 MHz
CPU4      4.0%        56 °C      3467 MHz
CPU5      2.0%        56 °C      3169 MHz
CPU6      0.0%        56 °C      4340 MHz
CPU7      0.0%        56 °C       400 MHz
CPU8      1.0%        56 °C       400 MHz
CPU9      1.0%        56 °C       400 MHz
CPU10      1.0%        56 °C       400 MHz
CPU11      0.0%        56 °C       400 MHz
CPU12      0.0%        56 °C      3210 MHz
CPU13      0.0%        56 °C      3453 MHz
CPU14      2.0%        56 °C      3452 MHz
CPU15      0.0%        56 °C       400 MHz

CPU fan speed: 0 RPM

auto-cpufreq version: 2.2.0 (git: )

Python: 3.10.14
psutil package: 5.9.8
platform package: 1.0.8
click package: 8.1.7
distro package: 1.9.0

Computer type: Notebook
Battery is: charging

auto-cpufreq system resource consumption:
cpu usage: 0.0 %
memory use: 0.09 %

Total CPU usage: 1.5 %
Total system load: 0.39
Average temp. of all cores: 56.00 °C 

Currently using: performance governor
Warning: CPU turbo is not available
Currently turbo boost is: off

-------------------------------------------------------------------------------
```

Issue #602, opened 2023-11-15
"Warning: CPU turbo is not available" - Of concern? 


gehuskic:
### Issue:
"Warning: CPU turbo is not available" is displayed in the GUI app. Also displays this in the "auto-cpufreq --stats" command. I am running am AMD CPU, so is turbo-boosting my CPU not an option for myself? 

- **More info**: 
- I tried using a config file with variable "turbo" set to "auto" for both power battery and charger governors (remembering to restart the daemon as well after telling the utility where the config is stored). Nothing change by doing so: 
`sudo auto-cpufreq --config /etc/auto-cpufreq.conf`
`sudo systemctl stop auto-cpufreq.service`
`sudo systemctl start auto-cpufreq.service`
-  Debug output and copy of the config file is all below. 

### Error output:

![image](https://github.com/AdnanHodzic/auto-cpufreq/assets/64159618/b4dbc31f-fb26-4abe-a450-a19f3cdae16e)

### System information:

<pre><font color="#26A269">gxnni@necrozama</font>:<font color="#26A269">/home/gxnni</font>$ sudo auto-cpufreq --debug

-------------------------------------------------------------------------------

Linux distro: Fedora Linux 39 Workstation Edition
Linux kernel: 6.5.6-300.fc39.x86_64
Processor: AMD Ryzen 5 7640U w/ Radeon 760M Graphics
Cores: 12
Architecture: x86_64
Driver: amd-pstate-epp

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 5579 MHz
CPU min frequency: 400 MHz

Core	Usage	Temperature	Frequency
CPU0      3.0%        36 °C      1376 MHz
CPU1     13.0%        36 °C      1735 MHz
CPU2      4.0%        36 °C      2392 MHz
CPU3      4.0%        36 °C       400 MHz
CPU4      1.0%        36 °C      1996 MHz
CPU5      3.0%        36 °C       400 MHz
CPU6      3.0%        36 °C       400 MHz
CPU7      1.0%        36 °C       400 MHz
CPU8      3.0%        36 °C       400 MHz
CPU9      9.0%        36 °C      1996 MHz
CPU10      4.0%        36 °C      1367 MHz
CPU11      0.0%        36 °C       400 MHz

auto-cpufreq version: 2.0.0 (git: 15c17fc)

Python: 3.12.0
psutil package: 5.9.6
platform package: 1.0.8
click package: 8.1.7
[...]

PR #674, opened 2024-04-07
gh-667: Modify message when using amd-pstate-epp

corona10:
Hi, First of all, auto-cpufreq is super useful for my laptop, thank you for working on this.

While I read the relevant threads, it looks like setting turbo can be delegated to the amd-pstate-epp.
Please let me know if I knew wrong :)

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -279,7 +279,7 @@ def turbo(value: bool = None):
         amd_value = amd_pstate.read_text().strip()
         if amd_value == "active":
             print("CPU turbo is controlled by amd-pstate-epp driver")
-        # Bascially, no other value should exist.
+        # Basically, no other value should exist.
         return False
     else:
         print("Warning: CPU turbo is not available")
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -276,8 +276,8 @@ def turbo(value: bool = None):
         f = cpufreq
         inverse = False
     elif amd_pstate.exists():
-        value = amd_pstate.read_text().strip()
-        if value == "active":
+        amd_value = amd_pstate.read_text().strip()
+        if amd_value == "active":
             print("CPU turbo is controlled by amd-pstate-epp driver")
             return False
         # Should not reach here.
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -279,8 +279,8 @@ def turbo(value: bool = None):
         amd_value = amd_pstate.read_text().strip()
         if amd_value == "active":
             print("CPU turbo is controlled by amd-pstate-epp driver")
-            return False
-        # Should not reach here.
+        # Bascially, no other value should exist.
+        return False
     else:
         print("Warning: CPU turbo is not available")
         return False
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -267,6 +267,8 @@ def turbo(value: bool = None):
     """
     p_state = Path("/sys/devices/system/cpu/intel_pstate/no_turbo")
     cpufreq = Path("/sys/devices/system/cpu/cpufreq/boost")
+    amd_pstate = Path("/sys/devices/system/cpu/amd_pstate/status")
+    scaling_driver = Path("/sys/devices/system/cpu/cpu0/cpufreq/scaling_driver")
 
     if p_state.exists():
         inverse = True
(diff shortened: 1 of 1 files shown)


corona10:
cc @LDPrg

corona10:
Screenshot from 2024-04-07 13-12-22

auto_cpufreq/core.py
```diff
@@ -267,13 +267,23 @@ def turbo(value: bool = None):
...
     elif cpufreq.exists():
         f = cpufreq
         inverse = False
+    elif amd_pstate.exists() and scaling_driver.exists():
+        amd_value = amd_pstate.read_text().strip()
+        scaling_driver_value = scaling_driver.read_text().strip()
+        if amd_value == "active" and scaling_driver_value == "amd-pstate-epp":
+            print("CPU turbo is controlled by amd-pstate-epp driver")
+        else:
+            print("Warning: CPU turbo is not available")
```
corona10, line 285: I am not sure if more status can exist at this moment which can set the turbo.
FYI, I am not the AMD expert :)

auto_cpufreq/core.py
```diff
@@ -267,13 +267,23 @@ def turbo(value: bool = None):
     """
     p_state = Path("/sys/devices/system/cpu/intel_pstate/no_turbo")
     cpufreq = Path("/sys/devices/system/cpu/cpufreq/boost")
+    amd_pstate = Path("/sys/devices/system/cpu/amd_pstate/status")
+    scaling_driver = Path("/sys/devices/system/cpu/cpu0/cpufreq/scaling_driver")
```
shadeyg56, line 271: Don't need to check `scaling_driver`. Just `amd_pstate/status` is enough

auto_cpufreq/core.py
```diff
@@ -267,13 +267,23 @@ def turbo(value: bool = None):
...
 
     if p_state.exists():
         inverse = True
         f = p_state
     elif cpufreq.exists():
         f = cpufreq
         inverse = False
+    elif amd_pstate.exists() and scaling_driver.exists():
```
shadeyg56, line 279: remove `scaling_driver.exists()`

auto_cpufreq/core.py
shadeyg56, line 285: These can be removed since this else statement will never be reached. the `cpufreq` Path does exist on `amd-pstate` so it would be caught by the preceding elif statement

auto_cpufreq/core.py
```diff
@@ -267,13 +267,23 @@ def turbo(value: bool = None):
...
+        amd_value = amd_pstate.read_text().strip()
+        scaling_driver_value = scaling_driver.read_text().strip()
+        if amd_value == "active" and scaling_driver_value == "amd-pstate-epp":
```
shadeyg56, line 282: remove the `scaling_driver` check since if `amd_pstate/status` equals "active", then the loaded driver must be `amd-pstate-epp`

shadeyg56: also this isn't really important but I think this entire if statement doesn't actually do anything since the `cpufreq` Path only doesn't exist on `amd-pstate-epp` but I think we should leave the if statement since it makes it more clear what is actually happening.

shadeyg56 (changes requested):
Left a few comments, but I think this is good in terms of making it more clear as to what is actually happening in the background

auto_cpufreq/core.py
corona10: hmm, by the way, what for amd_pstate == guided or amd_pstate == passive?

corona10: Self-reply for guided: https://github.com/AdnanHodzic/auto-cpufreq/issues/602#issuecomment-1945918261
/sys/devices/system/cpu/cpufreq/boost  is set.

shadeyg56:
Looks good, but still since `/sys/devices/system/cpu/cpufreq/boost` is set on both `amd_pstate=guided` and `amd_pstate=passive`, the else statement can be removed since this is handled by ` elif cpufreq.exists()`

Also i forgot to mention, the `return False` should be indented into the if statement

auto_cpufreq/core.py
```diff
@@ -267,13 +267,20 @@ def turbo(value: bool = None):
...
         f = p_state
     elif cpufreq.exists():
         f = cpufreq
         inverse = False
+    elif amd_pstate.exists():
+        value = amd_pstate.read_text().strip()
+        if value == "active":
+            print("CPU turbo is controlled by amd-pstate-epp driver")
+            return False
+        # Should not reach here.
```
corona10, line 283: I prefer to raise error here, but not sure about this project convention :)

corona10: Because following code will require `f`, but `f` is None at this moment.

shadeyg56: I get what you're saying. Its true but since it never reaches there, it shouldn't matter, but I suppose for future proofing, we can unindent the `return False`

shadeyg56 (approved):

shadeyg56:
Thanks for the changes. 

You'll be credited as part of the next release.

Issue #661, opened 2024-03-06
Intel CPU: Spamming of 'line 105: echo: write error: Device or resource busy


lamyergeier:
> Thank you for the PR, since these changes seem strictly AMD oriented, I've asked [on auto-cpufreq discord channel](https://discord.gg/Sjauxtj6kH) if someone could test these changes.
> 
> _Originally posted by @AdnanHodzic in https://github.com/AdnanHodzic/auto-cpufreq/issues/627#issuecomment-1902558770_

### I am still facing the issue but in Intel CPU.

Refer: [[Spamming of 'line 105: echo: write error: Device or resource busy' in journal · Issue #563 · AdnanHodzic/auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/issues/563)](https://github.com/AdnanHodzic/auto-cpufreq/issues/563)

![image](https://github.com/AdnanHodzic/auto-cpufreq/assets/42092626/a2514206-00db-4001-8db3-5dc25bcb377f)

```
$ systemctl status -o short-iso --full  auto-cpufreq
● auto-cpufreq.service - auto-cpufreq - Automatic CPU speed & power optimizer for Linux
     Loaded: loaded (/etc/systemd/system/auto-cpufreq.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: active (running) since Sun 2024-03-03 16:58:52 CET; 2 days ago
   Main PID: 1526 (python)
      Tasks: 1 (limit: 38105)
     Memory: 11.3M
        CPU: 1h 9min 28.718s
     CGroup: /system.slice/auto-cpufreq.service
             └─1526 /opt/auto-cpufreq/venv/bin/python /opt/auto-cpufreq/venv/bin/auto-cpufreq --daemon

2024-03-06T01:01:47+0100 lamyer python[4026913]: /usr/local/bin/cpufreqctl.auto-cpufreq: line 105: echo: write error: Device or resource busy
2024-03-06T01:01:47+0100 lamyer python[4026913]: /usr/local/bin/cpufreqctl.auto-cpufreq: line 105: echo: write error: Device or resource busy
2024-03-06T01:01:47+0100 lamyer python[4026913]: /usr/local/bin/cpufreqctl.auto-cpufreq: line 105: echo: write error: Device or resource busy
[...]

PR #680, opened 2024-04-18
Fix spam error message when setting energy_performance_preference

FosRexx:
The 'intel_pstate' driver does not allow the EPP to be set to anything but 'performance' when the scaling governor is set to 'performance', previously auto-cpufreq when the scaling-governor was set to 'performance' tried to set the EPP to 'balance-performance' which caused a spam of write error messages in journalctl in system with 'intel_pstate' drivers. This is an intended behavior, since according to the [kernel documentation](https://docs.kernel.org/admin-guide/pm/intel_pstate.html#hwp-performance) when HWP is enabled[(which is enabled by default during boot with supported processors)](https://docs.kernel.org/admin-guide/pm/intel_pstate.html#active-mode-with-hwp) and scaling governor is set to performance the processor’s internal P-state selection logic is expected to focus entirely on performance. And this will override the EPP setting and reject any value different from 0 (“performance”). This commit just changes the 'balance-performance' EPP preference in set_performance() to 'performance'. Which fixes the spam issue.

Fixes: #661

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -937,12 +937,20 @@ def set_performance():
         if dynboost_enabled:
             print('Not setting EPP (dynamic boosting is enabled)')
         else:
+            intel_pstate_status_path = "/sys/devices/system/cpu/intel_pstate/status"
+
             if conf.has_option("charger", "energy_performance_preference"):
                 epp = conf["charger"]["energy_performance_preference"]
+
+                if Path(intel_pstate_status_path).exists() and open(intel_pstate_status_path, 'r').read().strip() == "active" and epp != "performance":
+                    print(f'Warning "{epp}" EPP is not allowed in Intel CPU')
+                    print('Overriding EPP to "performance"')
+                    epp = "performance"
+
                 run(f"cpufreqctl.auto-cpufreq --epp --set={epp}", shell=True)
                 print(f'Setting to use: "{epp}" EPP')
             else:
-                if Path("/sys/devices/system/cpu/intel_pstate/status").exists() and open("/sys/devices/system/cpu/intel_pstate/status", 'r').read().strip() == "active":
+                if Path(intel_pstate_status_path).exists() and open(intel_pstate_status_path, 'r').read().strip() == "active":
                     run("cpufreqctl.auto-cpufreq --epp --set=performance", shell=True)
                     print('Setting to use: "performance" EPP')
                 else:
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -942,8 +942,12 @@ def set_performance():
                 run(f"cpufreqctl.auto-cpufreq --epp --set={epp}", shell=True)
                 print(f'Setting to use: "{epp}" EPP')
             else:
-                run("cpufreqctl.auto-cpufreq --epp --set=performance", shell=True)
-                print('Setting to use: "performance" EPP')
+                if Path("/sys/devices/system/cpu/intel_pstate/status").exists() is True:
+                    run("cpufreqctl.auto-cpufreq --epp --set=performance", shell=True)
+                    print('Setting to use: "performance" EPP')
+                else:
+                    run("cpufreqctl.auto-cpufreq --epp --set=balance_performance", shell=True)
+                    print('Setting to use: "balance_performance" EPP')
 
     # set frequencies
     set_frequencies()
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -122,14 +122,14 @@ except PermissionError:
     # Current work-around for Pop!_OS where symlink causes permission issues
     print("[!] Warning: Cannot get distro name")
     if os.path.exists("/etc/pop-os/os-release"):
-            # Check if using a Snap 
+            # Check if using a Snap
             if os.getenv("PKG_MARKER") == "SNAP":
-                print("[!] Snap install on PopOS detected, you must manually run the following" 
-                        " commands in another terminal:\n")            
+                print("[!] Snap install on PopOS detected, you must manually run the following"
+                        " commands in another terminal:\n")
                 print("[!] Backup the /etc/os-release file:")
-                print("sudo mv /etc/os-release /etc/os-release-backup\n")                
+                print("sudo mv /etc/os-release /etc/os-release-backup\n")
                 print("[!] Create hardlink to /etc/os-release:")
-                print("sudo ln /etc/pop-os/os-release /etc/os-release\n")            
+                print("sudo ln /etc/pop-os/os-release /etc/os-release\n")
                 print("[!] Aborting. Restart auto-cpufreq when you created the hardlink")
                 sys.exit(1)
             else:
@@ -182,7 +182,7 @@ def check_for_update():
             print("Error fetching recent release!")
             if message is not None and message.startswith("API rate limit exceeded"):
                 print("GitHub Rate limit exceeded. Please try again later within 1 hour or use different network/VPN.")
-            else: 
+            else:
                 print("Unexpected status code:", response.status_code)
             return False
     except (requests.exceptions.ConnectionError, requests.exceptions.Timeout,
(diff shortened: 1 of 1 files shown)


shadeyg56 (changes requested):
Thanks for finding the cause of the write error spam. but I think we want to keep `balance_performance` as the default EPP, so the solution to this would be to check if the user is running `intel_pstate`, if they are, set the EPP to `performance`, otherwise set it to `balance_performance`

Also does this not occur with `set_powersave()` as well since that uses `balance_power`? I can't test any of this as I don't have an Intel CPU

Lastly, in the future please try not to include all of that whitespace stuff in your commits

FosRexx:
In the original issue #661someone commented having the same issue with an AMD CPU with `amd-pstate-epp` driver and again in this closed issue #563 the same exact issue is reported for AMD CPU with `amd-pstate-epp` driver, which leads me to believe that this behavior happens in both AMD and Intel CPU with EPP support. I cannot confirm that this will work for the `amd-pstate-epp` drivers and I could not find anything in the [amd-pstate kernel documentation](https://docs.kernel.org/admin-guide/pm/amd-pstate.html) on this matter. So, I am hesitated to only apply this fix for `intel-pstate` drivers, since this could also fix the issue for `amd-pstate-epp` drivers, and the availability of EPP is already checked in the code before setting the `energy_performance_preference` so we don't have to worry about this applying for non-epp drivers. It would be great if someone using the `amd-pstate-epp` driver could verify this fix.

And yes this does not occur with `set_powersave()`, since in this mode the scaling governor is set to `powersave` and in this case their is no limitation to `energy_performance_perference` as stated [here](https://docs.kernel.org/admin-guide/pm/intel_pstate.html#hwp-powersave). This behavior at-least for the `intel-pstate` driver is intended and [mentioned](https://docs.kernel.org/admin-guide/pm/intel_pstate.html#hwp-performance) in the kernel documentation. The HWP feature is enabled by default for processors supporting it as mentioned [here](https://docs.kernel.org/admin-guide/pm/intel_pstate.html#active-mode-with-hwp).

FosRexx:
First, let me resolve the misunderstanding which was caused due to missing information in the PR comment and the commit message, which I have now hopefully resolved and new people stumbling on this PR will understand this better.

> Also, since I'm using intel_pstate driver am I misunderstanding something because when I'm on powersave governor, EPP will be set to power EPP?

So, [this](https://docs.kernel.org/admin-guide/pm/intel_pstate.html#hwp-performance) is the relevant section of the kernel documentation regarding the why's of the issue. In summary, in system's with Intel CPU and AMD CPU, that support the hardware-managed P-states (HWP) running the `intel_pstate` driver in active mode(which is the default), when the scaling governor is set to `performance`, the driver does not allow the `energy-performance-preference` to be set to any value different than `performance`. If the value is set to any value different than `performance` via `sysfs` interface, that configuration will be rejected.

This behavior causes the error message "write error: Device or resource busy", when auto-cpufreq tries to set the EPP to `balanced-performance` when the scaling governor is set to `performance` i.e. in `set_performance()` function.

This does not happen when the scaling governor is set to `powersave`, since in this case the driver allows the `energy-performance-preference` value to to be set to any value. That's why when you are on `powersave` governor, the EPP is set to `power`(which is weird, might I add, since in the latest commit 8bb7478 in master the EPP should be set to `balance_power` when in `powersave` governor, see function `set_powersave()`)

Hopefully this clears the misunderstanding.

Regarding keeping the default `balance_performance` as the default EPP.

> but I think we want to keep balance_performance as the default EPP, so the solution to this would be to check if the user is running intel_pstate, if they are, set the EPP to performance, otherwise set it to balance_performance

I am hesitated to only apply this fix for `intel_pstate`, because according to [this](https://github.com/AdnanHodzic/auto-cpufreq/issues/661#issuecomment-1986923978s) comment and currently closed issue #563 this issue also occurs in `amd_pstate_epp` driver. I do not have the first hand experience of this issue in AMD and I could not find anything regarding this on the `amd_pstate` [kernel documentation](https://docs.kernel.org/admin-guide/pm/amd-pstate.html), but as you can see from the comment and the Issue this also happens in AMD systems. And so, this PR might also fix the issue for `amd_pstate_epp` users, and thats why I am hesitant(I do not know if this expression makes any sense, as you might have already guessed English is not my first language). And so all the system with EPP are affected and non-EPP users should not be affected by this change in code, since the check for if the system has EPP or not is already been done.

shadeyg56:
> I am hesitated to only apply this fix for intel_pstate, because according to this comment and currently closed issue #563, this issue also occurs in amd_pstate_epp driver. I do not have the first hand experience of this issue in AMD and I could not find anything regarding this on the amd_pstate kernel documentation

I've used `amd_pstate_epp` before and can confirm that it supports balance_performance. Also based on  [these benchmarks](https://www.phoronix.com/review/linux-63-amd-epyc-epp), this shows amd can use balance_performance. I believe the issue you linked had a different cause for AMD systems. Currently, I'm not getting any error spam on my AMD system

FosRexx:
Well, I look like an idiot now 🥲, I already updated the code to only apply this for `intel_pstate` driver.

Parmjot-Singh:
> Also does this not occur with `set_powersave()` as well since that uses `balance_power`? I can't test any of this as I don't have an Intel CPU

Yes it does happen with Intel, I have tested it with the powersave governor and the power EPP has to be used with it. I think before merging, the MR should be changed to also include this case, but I do believe the way that this is done is not the best. I think that instead of just checking if the path exists, we should read the value of said path and set the value accordingly.

AdnanHodzic:
> I think before merging, the MR should be changed to also include this case, but I do believe the way that this is done is not the best. I think that instead of just checking if the path exists, we should read the value of said path and set the value accordingly.

+1

FosRexx:
> Yes it does happen with Intel, I have tested it with the powersave governor and the power EPP has to be used with it for the error to not be spammed in the logs.

Cannot reproduce
image

OS: Arch Linux x86_64
Host: Nitro AN515-58 (V2.10)
Kernel: 6.8.7-arch1-2
CPU: 12th Gen Intel(R) Core(TM) i5-12450H (12) @ 4.40 GHz

FosRexx:
Updated the code so that user written config energy_performance_preference are also handled.

AdnanHodzic:
> Updated the code so that user written config energy_performance_preference are also handled.

I started testing the changes on this branch, removed my auto-cpufreq config file and wanted to see what it runs on "default" settings. While laptop is connected to power it all works as expected in terms of EPP and governors.

But then, when pulled out the power cable and ran: `stress --cpu 8 --io 4 --vm 4 --timeout 15`

I noticed that both governor and EPP remained pinned to "powersave", and what's more worrying my CPU cores were also pinned to max 800Mhz. As while auto-cpufreq will switch you to `powersave` when when connected to the charger, it should switch over to `performance` if it detects such high cpu usage/load.

```
Linux distro: Ubuntu 24.04 Noble Numbat
Linux kernel: 6.8.0-31-generic
Cores: 8
Architecture: x86_64
Driver: intel_pstate

CPU max frequency: 800 MHz
CPU min frequency: 400 MHz

CPU0      2.1%        48 °C       800 MHz
CPU1     12.1%        48 °C       800 MHz
CPU2      8.2%        47 °C       765 MHz
CPU3     52.0%        49 °C       400 MHz
CPU4      3.0%        48 °C       400 MHz
CPU5      4.0%        48 °C       800 MHz
CPU6      0.0%        47 °C       800 MHz
CPU7     15.3%        49 °C       400 MHz

CPU fan speed: 0 RPM

Battery is: discharging

Setting to use: "power" EPP

Total CPU usage: 100 %
Total system load: 3.99
Average temp. of all cores: 48.00 °C 

Load optimal (load average: 3.99, 1.90, 1.37)
setting turbo boost: off
```

But then I noticed this was the case on `master` branch as well. Can anyone else reproduce this (i.e @shadeyg56 & @PurpleWazard), as if this is truly the case we had some kind of regression along the way.

AdnanHodzic:
Please disregard my last reply, from code POV, changes look good now, and I've been running auto-cpufreq with them for couple of days and I had 0 problems.

@shadeyg56 I requested re-review of comment I made, as from what I see this request has been fulfilled, if not please raise it.

Otherwise, thank you for your contribution @FosRexx, you will be credited for your work as part of upcoming auto-cpufreq release!

PR #691, opened 2024-05-09
More compatible for Arch based distributions

Angel-Karasu:
Change the method to detect if it's an Arch based distribution

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,46 +24,46 @@ Example of `auto-cpufreq --stats` CLI output
 
 ## Looking for developers and co-maintainers
 
-* If you would like to discuss anything regarding auto-cpufreq or its development, please join the [auto-cpufreq Discord server!](https://discord.gg/Sjauxtj6kH)
-* auto-cpufreq is looking for [co-maintainers & open source developers to help shape the future of the project!](https://github.com/AdnanHodzic/auto-cpufreq/discussions/312)
+- If you would like to discuss anything regarding auto-cpufreq or its development, please join the [auto-cpufreq Discord server!](https://discord.gg/Sjauxtj6kH)
+- auto-cpufreq is looking for [co-maintainers & open source developers to help shape the future of the project!](https://github.com/AdnanHodzic/auto-cpufreq/discussions/312)
 
 ## Index
 
-* [Why do I need auto-cpufreq?](#why-do-i-need-auto-cpufreq)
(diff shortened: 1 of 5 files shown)


AdnanHodzic:
Since I don't use Arch would like to hear what [auto-cpufreq AUR package maintainers](https://github.com/AdnanHodzic/auto-cpufreq/?tab=readme-ov-file#aur-package-archmanjaro-linux) thing about these changes.

@MusicalArtist12  @liljaylj and @Parmjot-Singh

Angel-Karasu:
All Arch based distributions use Arch repos (except Manjaro, but it's the same packages with a bit of a delay for better stability) so there shouldn't be any problem.

MusicalArtist12:
I mean, I don't see how it really changes, the original installer *does* work on arch linux, and it doesn't change the installer's logic in terms of what packages it installs to the system via `pacman`. I don't see anything wrong with it, I just think its superfluous. I also think the dinit support should be added through a separate pull request, since arch uses systemd anyways.

AdnanHodzic:
@Angel-Karasu could you then please also update Readme file as @MusicalArtist12 suggested, just so we avoid any ambiguity, thanks!

Angel-Karasu:
I don't understand what I should add in the README because if the AUR package works on Arch I don't see why it wouldn't work on Arch based distributions

MusicalArtist12:
I might've just misread the latest diff if that's the only change then. otherwise this looks good to me. I can separately create a PR that mentions that the AUR build doesn't use the installer.

Angel-Karasu:
The AUR package does not change and should normally already work on all Arch-based distributions. My change is to the installer, I'm just changing the way it detects the distribution, instead of using `/etc/os-release`, it checks if there is a `/etc/arch-release` file that is present on all Arch-based distributions.

AdnanHodzic:
> I might've just misread the latest diff if that's the only change then. otherwise this looks good to me. I can separately create a PR that mentions that the AUR build doesn't use the installer.

@MusicalArtist12 sounds good to me, separate PR for README update works for me. Maybe also a good idea to include/mention @Angel-Karasu under AUR/Arch section in case of questions, unless they are objections from @Angel-Karasu side.

@Angel-Karasu changes LGTM. Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #696, opened 2024-05-10
critical name error in auto-cpufreq --stats

yezarniko:


shadeyg56:
Thanks for catching this!

PR #701, opened 2024-05-14
Nix: fix duplicated package

shadeyg56:
I noticed when messing with my Nix config, that there were 2 versions of `auto-cpufreq` being installed, which didn't seem right. Took a look at our Nix module, and noticed it was pulling in the outdated version of `auto-cpufreq` from the `nixpkgs` repo alongside the git version

This is just a simple fix to stop it from pulling the unnecessary package and dependencies

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -31,15 +31,21 @@ in {
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


AdnanHodzic (approved):
LGTM

PR #702, opened 2024-05-16
Refactor charging function

dementive:
* Removed the `else` from the top level of the charging() function, the logic is the same but slightly easier to read now.

* Use `if not x.exists()` in the charging() function before opening the file instead of FileNotFoundError exceptions, makes the function a lot easier to read.

* Close the power_supply_type_path after reading the supply_type as it is not used later.

* Remove `else: continue` from the end of the charging() function for loop, this didn't actually do anything.

I tested the charging() function on my laptop plugged in and plugged out and it is still working as expected with these changes.

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -316,47 +316,41 @@ def charging():
     if len(power_supplies) == 0:
         # nothing found found, so nothing to check
         return True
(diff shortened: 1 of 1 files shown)


AdnanHodzic:

PR #709, opened 2024-05-21
Fixed auto-cpufreq error installation

wattanapong:
This error was encountered when the python was lower than 3.10. 
You need to install multiple python in /usr/bin/python3.10 where the default /usr/bin/python3.8 exists.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -112,7 +112,8 @@ function tool_install {
 
   if [ -f /etc/debian_version ]; then
     detected_distro "Debian based"
-    apt install python3-dev python3-pip python3-venv python3-setuptools dmidecode libgirepository1.0-dev libcairo2-dev libgtk-3-dev gcc -y
+    # apt install python3-dev python3-pip python3-venv python3-setuptools dmidecode libgirepository1.0-dev libcairo2-dev libgtk-3-dev gcc -y
+    apt install python3.10-dev python3-pip python3-venv python3-setuptools dmidecode libgirepository1.0-dev libcairo2-dev libgtk-3-dev gcc -y
 
   elif [ -f /etc/redhat-release ]; then
     detected_distro "RedHat based"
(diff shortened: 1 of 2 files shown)


auto-cpufreq-installer
```diff
@@ -125,12 +126,21 @@ function tool_install {
 
   header "Installing necessary Python packages"
 
+<<<<<<< HEAD
```
AdnanHodzic, line 129: Are you sure you want to merge this like this, it seems like you were in process of resolving a git commit?

AdnanHodzic, line 129: Are you sure you want to merge this like this, it seems like you were in process of resolving a git commit?

auto-cpufreq-installer
```diff
@@ -125,12 +126,21 @@ function tool_install {
...
+  venv_dir=/opt/auto-cpufreq/venv
+  mkdir -p "${venv_dir}"
+  python3.10 -m venv "${venv_dir}"
+
+  source "${venv_dir}/bin/activate"
+  python3.10 -m pip install --upgrade pip wheel
```
AdnanHodzic, line 135: Why would you want to limit it to 3.10 in particular, especially since latest version is 3.12 when otherwise it'll just use latest version?

auto-cpufreq-installer
```diff
@@ -125,12 +126,21 @@ function tool_install {
...
+=======
   venv_dir=$VENV_PATH/venv
   mkdir -p "$venv_dir"
   python3 -m venv "$venv_dir"
 
   source "$venv_dir/bin/activate"
   python3 -m pip install --upgrade pip wheel
+>>>>>>> 420648f5f12f6e5a486e71a4284b898a29f32459
```
AdnanHodzic, line 143: As I said above don't think these lines should be here ...

AdnanHodzic (changes requested):
Please address the comments I made.

AdnanHodzic:
Closed due to inactivity.

PR #718, opened 2024-06-01
nix: add overlay; use best practices

isabelroses:
Hey 👋 , been a while since I last checked up on this project.

Changes made:

- Remove deprecated [`lib.mdDoc`](https://github.com/NixOS/nixpkgs/blob/316b65cb7d266e3126314f566bd660decf6268af/lib/options.nix#L407)
- Add package overlay
- removed all instances of `with lib;` they are advised against, see https://github.com/NixOS/nixpkgs/issues/208242.
- `--replace` -> `--replace-fail`, using `--replace` will warn you that its being removed and the safe choice is `--replace-fail` since if something ever changes the build will fail.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1700823757,
-        "narHash": "sha256-PStAVLO8ycnFBSThYKTFwQ6rw+OH4ZM5KiN0fRE0OuM=",
+        "lastModified": 1717112898,
+        "narHash": "sha256-7R2ZvOnvd9h8fDd65p0JnB7wXfUvreox3xFdYWd1BnY=",
         "owner": "NixOS",
         "repo": "nixpkgs",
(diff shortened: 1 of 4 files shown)


isabelroses:
Due to the ci failing I updated the flake, seems you were running a pretty outdated nixpkgs input.
> Updated input 'nixpkgs':
    'github:NixOS/nixpkgs/068bacb9b6fbee6a603d1b1f68d7cf032c4feed1' (2023-11-24)
  → 'github:NixOS/nixpkgs/6132b0f6e344ce2fe34fc051b72fb46e34f668e0' (2024-05-30)

shadeyg56:
Awesome, I wasn't aware of most of that stuff so I appreciate this,

> Due to the ci failing I updated the flake, seems you were running a pretty outdated nixpkgs input.

Yeah I mean I expect most people to just use `nixpkgs.follows` so we don't really ever update our input unless there's a CI failure or we need a pin or whatever

cig0:
Btw thanks a ton to you both for maintaining this flake! <3

Issue #724, opened 2024-06-20
/usr/local/bin/cpufreqctl.auto-cpufreq: line 98: echo: write error: Device or resource busy


TheChilledBuffalo:
- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? **Yes**
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)? **No**
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)? **Yes**
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)? **No**

### Error output:
```text
sudo systemctl status auto-cpufreq.service          
[sudo] password for abhishek: 
● auto-cpufreq.service - auto-cpufreq - Automatic CPU speed & power optimizer for Linux
     Loaded: loaded (/etc/systemd/system/auto-cpufreq.service; enabled; preset: disabled)
     Active: active (running) since Thu 2024-06-20 12:59:25 IST; 7min ago
   Main PID: 602 (python)
      Tasks: 2 (limit: 18175)
     Memory: 40.3M (peak: 44.8M)
        CPU: 7.555s
     CGroup: /system.slice/auto-cpufreq.service
             └─602 /opt/auto-cpufreq/venv/bin/python /opt/auto-cpufreq/venv/bin/auto-cpufreq --daemon

Jun 20 13:07:10 archlinux-loq python[10949]: /usr/local/bin/cpufreqctl.auto-cpufreq: line 98: echo: write error: Device or resource busy
Jun 20 13:07:10 archlinux-loq python[10949]: /usr/local/bin/cpufreqctl.auto-cpufreq: line 98: echo: write error: Device or resource busy
Jun 20 13:07:10 archlinux-loq python[10949]: /usr/local/bin/cpufreqctl.auto-cpufreq: line 98: echo: write error: Device or resource busy
Jun 20 13:07:10 archlinux-loq python[10949]: /usr/local/bin/cpufreqctl.auto-cpufreq: line 98: echo: write error: Device or resource busy
[...]

PR #725, opened 2024-06-20
Use `performance` EPP in AMD CPUs when using `performance` governor

TheChilledBuffalo:
When the `performance` governor is being used, auto-cpufreq has been made to default to the `performance` EPP  when using `amd_pstate`.

All the details have been described in #724


AdnanHodzic:

PR #736, opened 2024-07-09
Code refactoring, more readable and easier to modify

Angel-Karasu:
###  Global variables
Reduce the number of processor calculations and easier to modify because only one change is needed instead of several

### Breaklines
Easier to modify because shorter code is more readable

### Empty `__init__.py` files
Empty so useless

### `--log` flag
The `--log` flag has been renamed to `--stats`

### `utils` folder to `config` folder
Easier to modify because more orderly folders

### Flags
Easier to modify because more orderly code

###  Imports and variables alphabetically
Easier to modify because more orderly and readable code

### `from pkg import objects` instead of `import pkg`
Easier to modify because more readable code


auto_cpufreq/battery_scripts/ideapad_laptop.py
```diff
@@ -17,14 +16,14 @@ def get_threshold_value(mode):
...
     try:
-        subprocess.check_output(f"echo {value} | tee /sys/bus/platform/drivers/ideapad_acpi/VPC2004:00/conservation_mode", shell=True, text=True)
+        subprocess.check_output(f"echo {value} | tee {CONSERVATION_MODE_FILE}", shell=True, text=True)
         print(f"conservation_mode is {value}")
     except: print("unable to set conservation mode")
     return
 
 def check_conservation_mode():
     try:
-        value = subprocess.check_output("cat /sys/bus/platform/drivers/ideapad_acpi/VPC2004:00/conservation_mode", shell=True, text=True)
```
AdnanHodzic, line 27: I actually like this change a lot, and I even suggested it in original PR.

auto_cpufreq/battery_scripts/ideapad_laptop.py
```diff
@@ -17,14 +16,14 @@ def get_threshold_value(mode):
...
+        value = subprocess.check_output(["cat", CONSERVATION_MODE_FILE], text=True)
```
AdnanHodzic, line 26: I actually like this change a lot, and I even suggested it in original PR.

auto_cpufreq/bin/auto_cpufreq.py
```diff
@@ -58,37 +48,20 @@ def config_info_dialog():
...
-                    countdown(2)
-                except KeyboardInterrupt: break
-            conf.notifier.stop()
-        elif monitor:
+        # set governor override unless None or invalid
+        if force is not None:
+            not_running_daemon_check()
+            root_check() # Calling root_check before set_override as it will require sudo access
+            set_override(force) # Calling set override, only if force has some values
+
```
AdnanHodzic, line 56: I fail to see how were all the lines on replaced with these new changes?  

Or was it moved to lines starting 113:

```
        elif daemon:
            config_info_dialog()
            root_check()
            file_stats()
            if IS_INSTALLED_WITH_SNAP and dcheck == "enabled":
                gnome_power_detect_snap()
                tlp_service_detect_snap()
            elif not IS_INSTALLED_WITH_SNAP:
                gnome_power_detect()
                tlp_service_detect()
            battery_setup()
            conf.notifier.start()
            while True:
                try:
                    footer()
                    gov_check()
                    cpufreqctl()
                    distro_info()
                    sysinfo()
                    set_autofreq()
                    countdown(2)
```

auto_cpufreq/bin/auto_cpufreq.py
```diff
@@ -115,7 +88,7 @@ def config_info_dialog():
             battery_setup()
             battery_get_thresholds()
             conf.notifier.start()
-            if os.getenv("PKG_MARKER") == "SNAP":
+            if IS_INSTALLED_WITH_SNAP:
```
AdnanHodzic, line 91: I also like what you did with global vars, as indeed this might be more readable.

auto_cpufreq/bin/auto_cpufreq.py
Angel-Karasu: The order has been changed to match the order of commands when you do `auto-cpufreq --help`.

auto_cpufreq/bin/auto_cpufreq.py
```diff
@@ -137,59 +110,32 @@ def config_info_dialog():
...
-            footer()
-        elif donate:
-            footer()
-            print("If auto-cpufreq helped you out and you find it useful ...\n")
-            print("Show your appreciation by donating!")
-            print("https://github.com/AdnanHodzic/auto-cpufreq/#donate")
-            footer()
+            battery_setup()
+            conf.notifier.start()
+            while True:
```
AdnanHodzic, line 125: Were all the changes on the left moved to some other lines? 

As I see big debug part being moved to i.e start 233:

```
        elif debug:
            # ToDo: add status of GNOME Power Profile service status
            config_info_dialog()
            root_check()
            battery_get_thresholds()
            cpufreqctl()
            footer()
            distro_info()
            sysinfo()
            print()
            app_version()
            print()
            python_info()
            print()
            device_info()
            print(f"Battery is: {'' if charging() else 'dis'}charging")
            print()
            app_res_use()
            get_load()
            get_current_gov()
            get_turbo()
            footer()
```

auto_cpufreq/bin/auto_cpufreq.py
```diff
@@ -231,15 +161,14 @@ def config_info_dialog():
...
+            if IS_INSTALLED_WITH_SNAP:
                 print("Detected auto-cpufreq was installed using snap")
                 # refresh snap directly using this command
                 # path wont work in this case
 
                 print("Please update using snap package manager, i.e: `sudo snap refresh auto-cpufreq`.")
                 #check for AUR 
-            elif subprocess.run(["bash", "-c", "command -v pacman >/dev/null 2>&1"]).returncode == 0 and subprocess.run(["bash", "-c", "pacman -Q auto-cpufreq >/dev/null 2>&1"]).returncode == 0:
-                print("Arch-based distribution with AUR support detected. Please refresh auto-cpufreq using your AUR helper.")
+            elif IS_INSTALLED_WITH_AUR: print("Arch-based distribution with AUR support detected. Please refresh auto-cpufreq using your AUR helper.")
```
AdnanHodzic, line 171: @shadeyg56 since IIRC you run Arch, can you confirm this line still works on your side?

auto_cpufreq/bin/auto_cpufreq_gtk.py
```diff
@@ -1,13 +1,8 @@
 #!/usr/bin/env python3
-
-import sys
-
-sys.path.append("../")
```
AdnanHodzic, line 5: How come this was not not needed at all?

auto_cpufreq/bin/auto_cpufreq.py
Angel-Karasu: The order has been changed to match the order of commands when you do `auto-cpufreq --help`

nix/patches/prevent-install-and-copy.patch
```diff
@@ -1,88 +1,73 @@
 diff --git a/auto_cpufreq/core.py b/auto_cpufreq/core.py
-index 99397a9..16869ab 100755
```
AdnanHodzic, line 2: @shadeyg56 if you could also check Nix part since I don't use it that would be great, thanks!

auto_cpufreq/bin/auto_cpufreq_gtk.py
Angel-Karasu: I don't know why it's not necessary at all, but as that is still working when I remove it, I'm removing it

auto_cpufreq/bin/auto_cpufreq.py
Angel-Karasu: I'm on Artix (Arch based) and it works to detect if `auto-cpufreq` has been installed with AUR

auto_cpufreq/bin/auto_cpufreq_gtk.py
Angel-Karasu: When one execute `auto-cpufreq` command, it calls `/opt/auto-cpufreq/venv/bin/auto-cpufreq` file, and it import `from auto_cpufreq.bin.auto_cpufreq import main` so it seems that the root folder is auto_cpufreq

auto_cpufreq/bin/auto_cpufreq.py
AdnanHodzic: Thought, that was the case, thanks for the clarification.

AdnanHodzic:
After thoroughly testing the changes, I'm happy to say LGTM! 

PR #738, opened 2024-07-15
Fix issue where epp cannot be set in `powersave` governor

TheChilledBuffalo:
Fixes error in #725 according to the latest comments in #724


JunaidQrysh (changes requested):
Add 'and gov == "performance":'
for intel as well

TheChilledBuffalo:
Oh right, I'll add that

TheChilledBuffalo:
Oops, the # 2 mention was an accident... was trying to say Number 2

AdnanHodzic:

PR #744, opened 2024-07-18
Fix AVAILABLE_GOVERNORS

rivenirvana:
The refactor in #736 missed a `strip()` from `get_avail_gov()` (see [blame](https://github.com/AdnanHodzic/auto-cpufreq/blame/34ebd04df0fd605c6c741f834f7b2d9205999f30/auto_cpufreq/core.py#L360)), which caused constant failures in `gov_check()` when trying to install the daemon, even if the required scaling governors are present.


AdnanHodzic:
Good catch! Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

Issue #737, opened 2024-07-13
Not detecting correct CPU Max frequency and ignoring config file


LDrawe:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? Yes
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)? N/A
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)? Yes
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)? Yes

### Error output:
```text
Add/paste error output in case of failed installation or other failing component
```
---

### System information:

Add/paste output of:

```
auto-cpufreq --debug
```

```
------------------------------ Current CPU stats ------------------------------

CPU max frequency: 3400 MHz
CPU min frequency: 400 MHz

Core    Usage   Temperature     Frequency
CPU0      1.0%        54 °C      3400 MHz
CPU1      1.0%        55 °C      3400 MHz
CPU2      4.0%        58 °C      3400 MHz
CPU3      3.0%        53 °C      3400 MHz
CPU4      3.1%        54 °C      3400 MHz
CPU5      6.1%        55 °C      3400 MHz
CPU6      2.0%        58 °C      3400 MHz
CPU7      2.0%        53 °C      3400 MHz

auto-cpufreq version: 2.3.0 (git: 432883a)

Python: 3.12.4
psutil package: 5.9.8
platform package: 1.0.8
click package: 8.1.7
distro package: 1.9.0

Computer type: Notebook
Battery is: charging

auto-cpufreq system resource consumption:
cpu usage: 0.0 %
memory use: 0.42 %

Total CPU usage: 3.5 %
Total system load: 0.85
Average temp. of all cores: 55.00 °C 

Currently using: performance governor
Currently turbo boost is: on

-------------------------------------------------------------------------------
```

Also please be descriptive about the issue you're reporting, i.e: what you tried & what's the expected behaviour.

[...]

PR #745, opened 2024-07-18
Fix: Getting incorrect max frequency value

LDrawe:
This PR (partially) closes https://github.com/AdnanHodzic/auto-cpufreq/issues/737. Instead of getting the max frequency from `cpuinfo_max_freq`, it gets from `scaling_max_freq`. It shoudn't be a problem for most systems, since these values tend to be the same. However, on some cases(like https://github.com/AdnanHodzic/auto-cpufreq/issues/737), they are reported different, with the `scaling_max_freq` being the correct one. Apparently this can happen if turbo boost is enabled by means other than the bios , like writing to the cpu registers directly or editing `/sys/devices/system/cpu/intel_pstate/no_turbo`

Any feedback about potential problems or improvements would be greatly appreciated.


LDrawe:
Closing this for now, as of yesterday I started getting mixed results, will investigate futher to see what's actually going on

PR #695, opened 2024-05-10
Minor refactor: code clean-up & make it more readable 

Angel-Karasu:
Make the code more readable

PR #749, opened 2024-07-22
Add some features for config

Angel-Karasu:
- Add detection of `$HOME/.config/auto-cpufreq.conf` file
- Add `--config-reload` option, the user chooses to enable or not the auto-reload when the configuration has been modified
- Auto install and remove `/etc/auto-cpufreq.conf` file
- Clean code
  - Add `has_option` and `get_option` functions
  - Remove `get_config` function because it's more cumbersome to call it

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -17,6 +17,7 @@ AUTO_CPUFREQ_FILE="/usr/local/bin/auto-cpufreq"
 AUTO_CPUFREQ_GTK_FILE=$AUTO_CPUFREQ_FILE-gtk
 AUTO_CPUFREQ_GTK_DESKTOP_FILE="$(basename $AUTO_CPUFREQ_GTK_FILE).desktop"
 
+CONFIG_FILE="/etc/auto-cpufreq.conf"
 IMG_FILE="/usr/share/pixmaps/auto-cpufreq.png"
 ORG_FILE="/usr/share/polkit-1/actions/org.auto-cpufreq.pkexec.policy"
 
@@ -140,6 +141,8 @@ function tool_install {
   mkdir -p $SHARE_DIR
   cp -r scripts/ $SHARE_DIR
   cp -r images/ $SHARE_DIR
+
+  cp "$(basename $CONFIG_FILE)-example" $CONFIG_FILE
   cp images/icon.png $IMG_FILE
   cp scripts/$(basename $ORG_FILE) $(dirname $ORG_FILE)
 
@@ -185,6 +188,7 @@ function tool_remove {
   remove_file $srv_remove
   remove_file $AUTO_CPUFREQ_FILE
   remove_file $AUTO_CPUFREQ_GTK_FILE
+  remove_file $CONFIG_FILE
   remove_file $IMG_FILE
   remove_file $ORG_FILE
   remove_file "/usr/local/bin/cpufreqctl.auto-cpufreq"
--- a/auto_cpufreq/battery_scripts/ideapad_acpi.py
+++ b/auto_cpufreq/battery_scripts/ideapad_acpi.py
@@ -2,8 +2,7 @@
 import os
 from subprocess import check_output
 
-from auto_cpufreq.config.config import config
-from auto_cpufreq.globals import POWER_SUPPLY_DIR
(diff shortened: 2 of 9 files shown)


AdnanHodzic:
> Add detection of $HOME/.config/auto-cpufreq.conf file

This is interesting to have, since there's an issue where user wants to use it from this location. Will most likely not work on Snap package installation, but regardless it could be feature that could only be used on non Snap installs.

I find `--reload` flag confusing, would rename it to `--config-reload` 

Also it's description should be updated from `Reload when the configuration has been modified` to something more descriptive, i.e: `Reload auto-cpufreq daemon to pick up changes made in auto-cpufreq config file` or something like that if that's what it's supposed to do?

Looking at changes, I also don't get how is this any different from approach @shadeyg56 did when he first implemented these changes? I made a comment on Discord, where I could only apply certain changes to [auto-cpufreq.conf file by restarting auto-cpufreq daemon with systemd.](https://discord.com/channels/1158239366251020361/1158239367337349164/1262381875952091136)

Also, currently this is what I get when I try to use reload flag, it appears nothing happens and terminal never clears out until I ctrl+c:
```
sudo auto-cpufreq --reload
Info: Using settings defined in /etc/auto-cpufreq.conf file
```

> Auto install and remove /etc/auto-cpufreq.conf file

How is this supposed to work, as I don't see it in code nor I see it explained anywhere?

Angel-Karasu:
> > Add detection of $HOME/.config/auto-cpufreq.conf file
> 
> This is interesting to have, since there's an issue where user wants to use it from this location. Will most likely not work on Snap package installation, but regardless it could be feature that could only be used on non Snap installs.
> 
> I find `--reload` flag confusing, would rename it to `--config-reload`
> 
> Also it's description should be updated from `Reload when the configuration has been modified` to something more descriptive, i.e: `Reload auto-cpufreq daemon to pick up changes made in auto-cpufreq config file` or something like that if that's what it's supposed to do?
> 
> Looking at changes, I also don't get how is this any different from approach @shadeyg56 did when he first implemented these changes? I made a comment on Discord, where I could only apply certain changes to [auto-cpufreq.conf file by restarting auto-cpufreq daemon with systemd.](https://discord.com/channels/1158239366251020361/1158239367337349164/1262381875952091136)
> 
> Also, currently this is what I get when I try to use reload flag, it appears nothing happens and terminal never clears out until I ctrl+c:
> 
> ```
> sudo auto-cpufreq --reload
> Info: Using settings defined in /etc/auto-cpufreq.conf file
> ```
> 
> > Auto install and remove /etc/auto-cpufreq.conf file
> 
> How is this supposed to work, as I don't see it in code nor I see it explained anywhere?

- `--reload` flag
- - Add an error message when used alone
- - Improves the description
- - Renames to `--config-reload`

Angel-Karasu:
Auto install auto-cpufreq.conf with `cp "$(basename $CONFIG_FILE)-example" $CONFIG_FILE`
Auto remove auto-cpufreq.conf with `remove_file $CONFIG_FILE`

Angel-Karasu:
I think `--config-reload` is useful because a lot of people change their config file once, it shouldn't be the default action because disabling it can increase security, reduce resource usage because there's no need to check the folder and as I said, in the majority of cases people don't touch the config file.

I removed the docstring because I found `find_config_file` very clear about what it does and I added a few comments to clarify the check order.

I didn't think of transforming the Config class into a ConfigManager child class, I think it's a great idea.

I said `$HOME` to be more concise but I didn't change the file detection method, it continues to use `getent passwd ${SUDO_USER:-$USER} | cut -d: -f6`

AdnanHodzic:
@Angel-Karasu as it was mentioned in another of your PR's, please create smaller PR's which are easier to review and test. I'm still [facing issues](https://discord.com/channels/1158239366251020361/1158239367337349164/1276809779003789342) from refactor introduced in #695 because there were so many changes which were hard to test individually. 

Since @shadeyg56 a person who introduced original "config reload" functionality doesn't understand the logic behind these changes (and neither do, my questions above has still not be answered).

As I mentioned in "[looking for maintainers and contributors](https://github.com/AdnanHodzic/auto-cpufreq/discussions/312#discussioncomment-9557496)" there's bunch of things that could be picked up to make auto-cpufreq better.

However, I don't see how these changes will make it better so I'm simply closing this PR.

Issue #755, opened 2024-07-31
Immediate crash in dinit service (+ FIX)


BowDown097:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted?  Y
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)? N
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)? Y
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)? Y

### Error output:
```text
Traceback (most recent call last):
  File "/usr/bin/auto-cpufreq", line 5, in <module>
    from auto_cpufreq.bin.auto_cpufreq import main
  File "/usr/lib/python3.12/site-packages/auto_cpufreq/bin/auto_cpufreq.py", line 14, in <module>
    from auto_cpufreq.core import *
  File "/usr/lib/python3.12/site-packages/auto_cpufreq/core.py", line 34, in <module>
    os.environ["PATH"] += ":/usr/bin"
    ~~~~~~~~~~^^^^^^^^
  File "<frozen os>", line 714, in __getitem__
KeyError: 'PATH'
```
---

### System information:

Add/paste output of:

```
auto-cpufreq --debug

Using settings defined in /etc/auto-cpufreq.conf file

-------------------------------- Battery Info ---------------------------------

battery count = 0

-------------------------------------------------------------------------------

Linux distro: Artix Linux rolling n/a
Linux kernel: 6.10.2-artix1-1
Processor: Intel(R) Core(TM) i7-4700MQ CPU @ 2.40GHz
Cores: 8
Architecture: x86_64
Driver: intel_cpufreq

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 2400 MHz
CPU min frequency: 800 MHz

Core	Usage	Temperature	Frequency
CPU0      1.0%        71 °C      1311 MHz
CPU1      1.0%        71 °C       800 MHz
CPU2      1.0%        71 °C      2299 MHz
CPU3      0.0%        71 °C       800 MHz
[...]

PR #756, opened 2024-07-31
Fix unsafe access to PATH

BowDown097:
This addresses #755 - an immediate crash due to a KeyError when PATH is not present in os.environ (which is the case when the service is ran through dinit).


AdnanHodzic:
LGTM, Thank you for your contribution, you will be credited for your work as part of [future release.](https://github.com/AdnanHodzic/auto-cpufreq/releases)

Issue #707, opened 2024-05-21
Nix package broken after requests update


Svenum:
The Commit https://github.com/AdnanHodzic/auto-cpufreq/commit/2b09bf7543ee181c5e5b30358cef0e1603d5fd79 broke the nix package. On Nix the newest verison of requests is 2.31.0.

shadeyg56:
Can you provide more information because it looks like the checks passed. What exactly is happening? A bump in the `poetry.lock` shouldn't affect the Nix flake

Svenum:
```console
warning: input 'nixos-hardware' has an override for a non-existent input 'nixpkgs'                                                                                                                                                                                                                                                                   
building the system configuration...                                                                                                                                                                                                                                                                                                                 
warning: input 'nixos-hardware' has an override for a non-existent input 'nixpkgs'                                                                                                                                                                                                                                                                   
error: builder for '/nix/store/2b07kvhzgdx0fc4gzz09ixfvimmkpgy1-python3.11-auto-cpufreq-2.3.0.drv' failed with exit code 1;                                                                                                                                                                                                                          
       last 25 log lines:                                                                                                                                                                                                                                                                                                                            
       > source root is alwf5z8m8v2jyqvhnf42889v6qdc6dc7-source                                                                                                                                                                                                                                                                                      
       > setting SOURCE_DATE_EPOCH to timestamp 315619200 of file alwf5z8m8v2jyqvhnf42889v6qdc6dc7-source/snap/snapcraft.yaml                                                                                                                                                                                                                        
       > Running phase: patchPhase                                                                                                                                                                                                                                                                                                                   
       > applying patch /nix/store/nlmyiyrdz7klcflh72rvd45njmmi4zbl-prevent-install-and-copy.patch                                                                                                                                                                                                                                                   
       > patching file auto_cpufreq/core.py                                                                                                                                                                                                                                                                                                          
       > Hunk #1 succeeded at 391 (offset 41 lines).                                                                                                                                                                                                                                                                                                 
       > Hunk #2 succeeded at 424 (offset 41 lines).                                                                                                                                                                                                                                                                                                 
       > Hunk #3 succeeded at 465 (offset 41 lines).                                                                                                                                                                                                                                                                                                 
       > substituteStream(): WARNING: '--replace' is deprecated, use --replace-{fail,warn,quiet}. (file 'auto_cpufreq/core.py')                                                                                                                                                                                                                      
[...]

Issue #654, opened 2024-02-22
Disabling CPU cores crashes the daemon.


shadeyg56:
This bug was discovered in #603. 
Running `echo 0 | sudo tee /sys/devices/system/cpu/cpuN/online` (where N is any integer greater than 0) causes the daemon to crash

This can be reproduced by running the command then running `systemctl status auto-cpufreq` or `auto-cpufreq --stats`

I'm going to look into this and hopefully open a PR to resolve this issue.

shadeyg56:
Unfortunately, this looks like an upstream bug in `psutil`
https://github.com/giampaolo/psutil/issues/2254

So this isn't something we can fix. The issue on the psutil repo has had little activity as well. If I have some time, I might delve into the psutil source code and see if I can open a PR over there

shadeyg56:
Ok i've submitted a PR on the psutil repo that should fix this.
https://github.com/giampaolo/psutil/pull/2376

I imagine it will take some time for this to merge and make it into a release so we just have to wait.

AdnanHodzic:
Congrats on getting it merged!

astonish-g:
Thank you so much beeing nice enough to fix this issue! Now I changed my laptop, and have one with AMD Ryzen 8845HS. I will try this, but for now I am getting higher watt useage compared to my old one. I am still trying to find the issue, before I will try this fix. Thank you again!

Svenum:
Hi,
I do not know if this is the right place, but I do not want to open an new issue only for this question. 
Is there a way to set CPU-Core-Count via the config file? I want to set my core count to 4 if on battery and to 16 if on AC.

shadeyg56:
Its not currently possible. 
1. because it hasn't been implemented
2. this bug is still in upstream as the merged PR has not made its way into a release yet so disabling a cpu core will crash auto-cpufreq

Nolicx:
Hey there, i've been using this tool for about the past week and i think it's great.
While searching for a way to disable cores i've found this issue and seen a few other issues and discussions regariding this feature.
I would love to implement it, as soon as the upstream issue has been fixed. 
Is there anything i should do besides the implementation and the creation of a pull request?

shadeyg56:
I'm glad auto-cpufreq has been useful for you.

If you're interested in adding it, I think its a good idea and shouldn't be too difficult.
[...]

PR #759, opened 2024-08-03
Unpin psutil and requests packages

shadeyg56:
We had these packages pinned because of #654 and #707. 
The psutil package had an official update so we no longer need to pull from the github repo and the requests package got updated in Nix so its now safe to unpin both of these dependencies

Also added poetry to the dev shell for Nix for convenience


AdnanHodzic:
LGTM and makes total sense! Thanks

Issue #522, opened 2023-06-13
Make use of `energy_perf_bias`?


ret2src:
Hi,

first of all: Thank you for creating `auto-cpufreq`, I really like it and prefer it to more invasive alternatives like TLP.

I'm fairly new to Linux power management and just found out about the `energy_perf_bias` value for Intel processors (see <https://wiki.archlinux.org/title/CPU_frequency_scaling#Intel_performance_and_energy_bias_hint>).

Currently, it is set to the default value of `6` on my system although `auto-cpufreq` set the governor to `performance`.
I wonder if setting the value to `0` has the potential to further improve performance? If so, would it make sense to include such a feature in `auto-cpufreq`?

AdnanHodzic:
Could be an interesting thing to have ... since I'm very limited with available time to do this myself, if you or anyone else wants to give it a try in implementing this functionality, please give it a try and [contribute to the project](https://github.com/AdnanHodzic/auto-cpufreq/#code-contribution) and you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

AdnanHodzic:
This functionality has been implemented as part of #771, please refer to [config file ](https://github.com/AdnanHodzic/auto-cpufreq?tab=readme-ov-file#3-auto-cpufreq-config-file)on how to enable & set it up.

ret2src:
Awesome, thank you!

PR #771, opened 2024-09-08
Add EPB (Energy Performance Bias) settings support

snyypi:
This PR adds support for the `energy_perf_bias` setting, giving users more control over how their Intel CPU (`intel_pstate` driver) balances performance and power efficiency

What's Changed:
- Introduced a new `energy_perf_bias` option, which can be set as a number (0-15) or using predefined strings like "performance", "balance_performance", "default", "balance_power" and "power" (made their names the same as EPP's)
- Integrated the `energy_perf_bias` setting into the existing power management logic, dynamically adjusting it based on the selected profile or user preferences
- Updated the documentation to explain how to configure and use the new `energy_perf_bias` setting

Closes: #522 

Looking forward to any feedback or suggestions!

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,14 @@ governor = performance
 # EPP: see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
 energy_performance_preference = performance
 
+# EPB (Energy Performance Bias) for the intel_pstate driver
+# see conversion info: https://www.kernel.org/doc/html/latest/admin-guide/pm/intel_epb.html
+# available EPB options include a numeric value between 0-15
+# (where 0 = maximum performance and 15 = maximum power saving),
+# or one of the following strings:
+# performance (0), balance_performance (4), default (6), balance_power (8), or power (15)
+# energy_perf_bias = balance_performance
+
 # Platform Profiles
 # https://www.kernel.org/doc/html/latest/userspace-api/sysfs-platform_profile.html
 # See available options by running:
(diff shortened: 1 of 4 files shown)


AdnanHodzic:
LGTM & this is great stuff! Making my experience much smoother.

It would be great if it could be enabled by default (automatically) without having to explicitly specify it as part of config file.

Setting i.e `energy_perf_bias = balance_power` when on powersavings governor & `energy_perf_bias = balance_performance` when on performance? 

Either way, thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

snyypi:
@AdnanHodzic Thank you for the amazing project! Regarding the configuration, if the parameter is missing in the configuration file, `balance_performance` (on charge) and `balance_power` (on battery) will be used if the hardware supports it. My mistake, I forgot to specify this in both the PR description and the documentation

AdnanHodzic:
> Regarding the configuration, if the parameter is missing in the configuration file, balance_performance (on charge) and balance_power (on battery) will be used if the hardware supports it. 

Yes I saw that in lines like

```
    if Path("/sys/devices/system/cpu/intel_pstate").exists() is False:
        print('Not setting EPB (not supported by system)')
        return
    epb = "balance_performance" if profile == "charger" else "balance_power"
```

But I misinterpreted it that it would be used only as part of config file if you didn't explicitly define other settings.

> My mistake, I forgot to specify this in both the PR description and the documentation

Could you please create a PR to update ReadMe with this info, just to avoid any confusion? Thanks!

PR #772, opened 2024-09-09
Fixed wrong sensor being monitored (AMD)

johanssonvincent:
When I installed auto-cpufreq on my fresh EndevourOS install it was monitoring the wrong sensor, and showing temperatures of 17 degrees. 

This is the fix I did, and it is currently showing the correct temp:
Instead of only looking for 'CPU' temp.label, I added that it should look for 'Tctl', as that is the sensor for average temp on my 7800X3D CPU.


AdnanHodzic:

AdnanHodzic:

PR #788, opened 2024-11-03
Improve TuneD detection mechanism

geragio:
This PR improves the detection mechanism for TuneD on the host system. 
Previously, when TuneD was removed or absent, the system would generate an error when removing auto-cpufreq deamon or when starting the software in live mode:
~~~
* Enabling TuneD

Unit tuned.service does not exist, proceeding anyway.
Failed to enable unit: Unit file tuned.service does not exist.`
~~~

With this improvement, auto-cpufreq now gracefully handles the absence of TuneD without producing misleading errors.


AdnanHodzic:
Great! Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

Issue #740, opened 2024-07-17
[Support Request] Fail to start auto-cpufreq in NixOS with flake installation


piyoki:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? 
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)?
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)?
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)?

### Error output:

```text
Add/paste error output in case of failed installation or other failing component
```

See the systemd logs below.

---

### System information:

Add/paste output of:

```bash
auto-cpufreq --debug
```

```console
-------------------------------- Battery Info ---------------------------------

battery count = 1
BAT0 start threshold = 0
BAT0 start threshold = 99

-------------------------------------------------------------------------------

Linux distro: NixOS 24.11 Vicuna
Linux kernel: 6.9.9-cachyos
Processor: Intel(R) Core(TM) Ultra 7 155H
Cores: 22
Architecture: x86_64
Driver: intel_pstate

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 4800 MHz
CPU min frequency: 400 MHz

Core	Usage	Temperature	Frequency
CPU0      0.0%        50 °C      3343 MHz
CPU1      0.0%        50 °C       400 MHz
CPU2      0.0%        44 °C       400 MHz
CPU3      0.0%        44 °C       400 MHz
CPU4      0.0%        45 °C       400 MHz
CPU5      0.0%        45 °C       400 MHz
CPU6      1.0%        44 °C       400 MHz
CPU7      1.0%        44 °C       400 MHz
CPU8      0.0%        45 °C      1665 MHz
CPU9      0.0%        45 °C       400 MHz
CPU10      0.0%        46 °C      1958 MHz
CPU11      0.0%        46 °C       400 MHz
CPU12      0.0%        46 °C       400 MHz
CPU13      0.0%        45 °C       400 MHz
CPU14      0.0%        46 °C       400 MHz
[...]

PR #809, opened 2025-02-09
nix: fail evaluation if power-profiles-daemon is enabled

shadeyg56:
So I noticed auto-cpufreq systemd service was being killed immediately after boot on Nix. 
This was actually brought up previously in #740, but it seemed to fix itself for everyone.
Anyways, it started happening to me recently and was happening on every single boot so I knew something was wrong.

Basically I discovered that if a module enabled (or the user has it explicitly enabled)  `services.power-profiles-daemon`, systemd sees it as a conflict and just decides to kill auto-cpufreq in favor of power-profiles. In my situation, Cinnamon DE was enabling it by default.

This PR makes a change that causes an assertion error to be thrown if the user has `auto-cpufreq` enabled alongside `power-profiles`. This forces the user to disable `power-profiles`, in order for their config to build. I considered just making this a warning, since technically you can still run `auto-cpufreq` manually, but I feel like it is better to force users to enable one or the other in this case

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -50,5 +50,15 @@ in {
         ];
       };
     };
+
+    assertions = [
+      {
+        assertion = !config.services.power-profiles-daemon.enable;
+        message = ''
+          You have set services.power-profiles-daemon.enable = true;
+          which conflicts with auto-cpufreq
+        '';
+      }
+    ];
   };
 }


AdnanHodzic:
LGTM, wouldn't be bad to have this same [`assertions` block](https://github.com/AdnanHodzic/auto-cpufreq/pull/809/files) outside of Nix scope.

Issue #824, opened 2025-03-26
GUI not open on Linux mint


sibbirAhmad:
When I click on the GUI icon, nothing happens

AdnanHodzic:
If you want me or any of the other maintainers to help you, do not  ignore issue template! Which you did, hence making this issue invalid, closing. Feel free to re-open this issue once you properly filled out requested fields.

Reference:

### Fill out information requested in this template, without doing so issue will be ignored & closed!

#### Before submitting an issue, it is strongly recommended to use the **[auto-cpufreq-genAI-chatbot](https://foolcontrol.org/?p=4903)** to get an immediate answer to your question.

### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? 
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)?
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)?
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)?

### Error output:
```text
Add/paste error output in case of failed installation or other failing component
```
---

### System information:

Add/paste output of:

```
auto-cpufreq --debug
```

Also please be descriptive about the issue you're reporting, i.e: what you tried & what's the expected behavior.

kirmola:
refer #825

Issue #825, opened 2025-03-28
GUI not launching on linux mint Cinnamon edition


kirmola:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? 
 ### YES, one issue related to this was closed for not filling this template
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)?
### Not Applicable
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)?
### Yes, N/A
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)?
### Error output:
```text
kirmola@kirmola-IdeaPad-5-15ITL05:~$ sudo auto-cpufreq-gtk 
Traceback (most recent call last):
  File "/opt/auto-cpufreq/venv/bin/auto-cpufreq-gtk", line 5, in <module>
    from auto_cpufreq.bin.auto_cpufreq_gtk import main
  File "/opt/auto-cpufreq/venv/lib/python3.12/site-packages/auto_cpufreq/bin/auto_cpufreq_gtk.py", line 2, in <module>
    import gi
ModuleNotFoundError: No module named 'gi'

```
---

### System information:

Add/paste output of:

```
auto-cpufreq --debug
```
```

-------------------------------- Battery Info ---------------------------------

battery count = 1
cat: /sys/class/power_supply/BAT0/charge_start_threshold: No such file or directory
ERROR: failed to read battery BAT0 thresholds: CalledProcessError(1, ['cat', '/sys/class/power_supply/BAT0/charge_start_threshold'])

-------------------------------------------------------------------------------

Linux distro: Linux Mint 22.1 Xia
Linux kernel: 6.8.0-56-generic
Processor: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
Cores: 8
Architecture: x86_64
Driver: intel_pstate

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 2400 MHz
CPU min frequency: 400 MHz

Core	Usage	Temperature	Frequency
[...]

Issue #813, opened 2025-03-10
Dependency 'girepository-2.0' is required but not found.


sergeolkhovik:
Hi! Have Carbon X1 gen 6 laptop running Ubuntu 24.10 and get this error while trying to install according to the doc:

```
Collecting PyGObject<4.0.0,>=3.46.0 (from auto-cpufreq==2.5.0+6be1547)
  Downloading pygobject-3.52.1.tar.gz (1.2 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.2/1.2 MB 7.8 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [23 lines of output]
      + meson setup /tmp/pip-install-d1gea8ab/pygobject_6192d5ef207c4a78b56f0eb36df0f755 /tmp/pip-install-d1gea8ab/pygobject_6192d5ef207c4a78b56f0eb36df0f755/.mesonpy-77k4i7a0 -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Dtests=false -Dwheel=true --wrap-mode=nofallback --native-file=/tmp/pip-install-d1gea8ab/pygobject_6192d5ef207c4a78b56f0eb36df0f755/.mesonpy-77k4i7a0/meson-python-native-file.ini
      The Meson build system
      Version: 1.7.0
      Source dir: /tmp/pip-install-d1gea8ab/pygobject_6192d5ef207c4a78b56f0eb36df0f755
      Build dir: /tmp/pip-install-d1gea8ab/pygobject_6192d5ef207c4a78b56f0eb36df0f755/.mesonpy-77k4i7a0
      Build type: native build
      Project name: pygobject
      Project version: 3.52.1
      C compiler for the host machine: cc (gcc 14.2.0 "cc (Ubuntu 14.2.0-4ubuntu2) 14.2.0")
      C linker for the host machine: cc ld.bfd 2.43.1
      Host machine cpu family: x86_64
      Host machine cpu: x86_64
      Program python3 found: YES (/opt/auto-cpufreq/venv/bin/python)
[...]

PR #826, opened 2025-03-28
A Fix for GUI not launching in some cases.

kirmola:
libgirepository-2.0-dev is now required for PyGObject which is required for gi module in [ python file which initiate GUI](https://github.com/AdnanHodzic/auto-cpufreq/blob/master/auto_cpufreq/gui/app.py) see: https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/320

so here is sample PR for issue #825

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -76,26 +76,18 @@ function tool_install {
     header "Setting up Python environment"
   }
 
-  if [ -f /etc/debian_version ]; then
-    detected_distro "Debian based"
+  if [ -f /etc/debian_version ]; then  
+    VERSION=$(cat /etc/debian_version)  
 
-    # Load system info
-    . /etc/os-release  
-
-    # Extract major version number (e.g., "22.1" → "22")
-    VER=${VERSION_ID%%.*}
-
-    # Default to libgirepository-2.0-dev
-    LIB_GI_REPO="libgirepository-2.0-dev"
-
-    # Use the older package for Debian < 13 and Ubuntu < 24.04
-    if [[ "$VER" -lt 13 && "$VERSION_CODENAME" != "trixie" && "$VERSION_CODENAME" != "sid" ]]; then
+    if [[ "$VERSION" -ge 13 ]]; then  
+        LIB_GI_REPO="libgirepository-2.0-dev"
+    else  
         LIB_GI_REPO="libgirepository1.0-dev"
     fi
 
-    # Install required packages
+    
     apt install -y python3-dev python3-pip python3-venv python3-setuptools dmidecode \
-        "$LIB_GI_REPO" libcairo2-dev libgtk-3-dev gcc
+    "$LIB_GI_REPO" libcairo2-dev libgtk-3-dev gcc
 
   elif [ -f /etc/redhat-release ]; then
     detected_distro "RedHat based"
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -86,11 +86,7 @@ function tool_install {
     fi
 
     echo $LIB_GI_REPO needs to be installed for version $VERSION
-    echo =====================================
-    echo =====================================
-    echo =====================================
-    echo =====================================
-    echo =====================================
+    echo '---- '
     apt install -y python3-dev python3-pip python3-venv python3-setuptools dmidecode \
     "$LIB_GI_REPO" libcairo2-dev libgtk-3-dev gcc
 
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -78,14 +78,19 @@ function tool_install {
 
   if [ -f /etc/debian_version ]; then  
     VERSION=$(cat /etc/debian_version)  
-
-    if [[ "$VERSION" -ge 13 ]]; then  
-        LIB_GI_REPO="libgirepository-2.0-dev"
-    else  
+    # VERSION="CSCS"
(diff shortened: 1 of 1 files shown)


auto-cpufreq-installer
```diff
@@ -78,7 +78,7 @@ function tool_install {
 
   if [ -f /etc/debian_version ]; then
```
AdnanHodzic, line 79: Since for anyone (or any distro derivative) that's using current (stable) version of Debian 12 or anything lower then Ubuntu 24.04 ([as I mentioned here](https://github.com/AdnanHodzic/auto-cpufreq/issues/813#issuecomment-2726246869)), let's extend this logic/loop so that it installs `libgirepository1.0-dev` in case of older releases otherwise go with `libgirepository-2.0-dev` for anything Debian >= 13 (trixie) and let's merge this.

AdnanHodzic:
Note to myself for release notes, this issue fixes: #824, #825 &  #813

auto-cpufreq-installer
```diff
@@ -78,7 +78,24 @@ function tool_install {
...
+    # Load system info
+    . /etc/os-release  
+
+    # Extract major version number (e.g., "22.1" → "22")
+    VER=${VERSION_ID%%.*}
+
+    # Default to libgirepository-2.0-dev
+    LIB_GI_REPO="libgirepository-2.0-dev"
+
+    # Use the older package for Debian < 13 and Ubuntu < 24.04
```
AdnanHodzic, line 91: Let's do it differently, something like:

```
if [ -f /etc/debian_version ]; then
    VERSION=$(cat /etc/debian_version)
    if ! [[ $VERSION =~ ^12(\.[0-9]+)?$ ]]; then
        # The Debian version is not 12 or any 12.X version
        # go with libgirepository-2.0-dev
    else
        # The Debian version is 12 or a 12.X version
        # go with libgirepository1.0-dev
    fi
fi
```

Because instead of using codenames, Debian uses release numbers for stable releases. So if it's anything lower then 12.xx then it should go with `libgirepository1.0-dev`

Hence why `cat /etc/debian_version` on Ubuntu >= 24.04 (and all of its derivatives) should already returns `trixie/sid`. Trixie = testing (future 13), and Sid = unstable, never to be released.

AdnanHodzic: I don't think you understood why I was suggesting what I suggested. Please re-read:

> .... Because instead of using codenames, **Debian uses release numbers for stable releases.** So if it's anything lower ...

So using `if [[ "$VERSION" -ge 13 ]]; then` will not work, because 13 doesn't exist yet so this statement will always return **false** until "trixie" is turned into 13. That's why I suggested:

```
if [ -f /etc/debian_version ]; then
    VERSION=$(cat /etc/debian_version)
    if ! [[ $VERSION =~ ^12(\.[0-9]+)?$ ]]; then
        # The Debian version is not 12 or any 12.X version
        # go with libgirepository-2.0-dev
    else
        # The Debian version is 12 or a 12.X version
        # go with libgirepository1.0-dev
    fi
fi
```

So in this case if you ran code above, and your `cat /etc/debian_version` return `whatever` it would go with never as it'll catch case of any future codenames. 

If you tested code you pushed you would see that it also returns an error:

```
sudo ./auto-cpufreq-installer --install

./auto-cpufreq-installer: line 82: [[: trixie/sid: division by 0 (error token is "sid")
```

AdnanHodzic:
Are there [any updates](https://github.com/AdnanHodzic/auto-cpufreq/pull/826#discussion_r2019705078)? Would be great to have this merged.

kirmola:
> Are there [any updates](https://github.com/AdnanHodzic/auto-cpufreq/pull/826#discussion_r2019705078)? Would be great to have this merged.

yeah, pushed changes, please look

auto-cpufreq-installer
```diff
@@ -78,14 +78,19 @@ function tool_install {
...
+    # VERSION="CSCS"
+    if [[ "$VERSION" =~ ^12(\.[0-9]+)?$ || ("$VERSION" =~ ^[0-9]+$ && "$VERSION" -lt 12) ]]; then  
         LIB_GI_REPO="libgirepository1.0-dev"
+    else  
+        LIB_GI_REPO="libgirepository-2.0-dev"
     fi
 
-    
+    echo $LIB_GI_REPO needs to be installed for version $VERSION
+    echo =====================================
```
AdnanHodzic, line 89: One final remark, let's change lines 89-93 to a single:
```
echo " "
```

or 

```
echo "---- "
```

AdnanHodzic:
Great stuff! Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

kirmola:
All good, you are maintaining an excellent package, keep doing great stuff.

AdnanHodzic:
> All good, you are maintaining an excellent package, keep doing great stuff.

Thanks and will do. Needless to say thank you and all other contributors who make this project what it is.

Btw, with your PR merged, [we've hit 100 contributor point](https://github.com/AdnanHodzic/auto-cpufreq/graphs/contributors), so congrats for being the 100! :smile:

kirmola:
> 100

its special 🥳

AdnanHodzic:
New issues coming out of this implementation, #833, #834,

AdnanHodzic:
Pushed a fix as part of [8870b62 commit](https://github.com/AdnanHodzic/auto-cpufreq/commit/8870b62b172b11fea9de63971158a76c271d939a)

AdnanHodzic:
Made another update where I updated logic to determine the `PyGObject` version based on the corresponding `libgirepository` apt package version and update it in pyproject.toml.

PR #838, opened 2025-04-21
ideapad_fixed_threshold_manufacturer

Genius-19:


AdnanHodzic:
I would appreciate if PR description would have a brief (even single sentence is good) description of what a PR does.

Regarding following sentences:

> set by the manufacturer
& 
> value set in the system by the manufacturer.

Is there a way user can see to what values are set by the manufacturer, if yes I think it would be good idea to add that info. 

@PurpleWazard since you're original implementer of this functionality, any comments from your side before this is merged?

Genius-19 (commented):
This is a specific config concerning model laptops with battery threshold limit value set by the manufacturer.

PurpleWazard:
@AdnanHodzic I see nothing wrong with this PR. Since I don't have a laptop from Lenovo with this I can't test. @Genius-19 Thanks for your PR your guide should help many.

Genius-19:
I've tested for a week on my lenovo ideapad including all the classic tests (plug, unplug, reboot, change values, etc).
Thank you all for this amazing software

AdnanHodzic:
>Thank you all for this amazing software

PR #841, opened 2025-05-02
fix: rectify window icon under Wayland

Integral-Tech:
Set the correct desktop filename to fix window icon under Wayland.


AdnanHodzic:
Great one, LGTM! Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #855, opened 2025-06-21
Fix `auto-cpufreq --monitor` reporting the "CPU frequency scaling" section wrong.

PhoenixCausesOof:
It used to always read the config file's "battery" section and report that, even if connected to a charger. I used a hacky solution (simply check the state of the battery, i.e., charging or not charging) and read the correct part of the config file based on that.

### DIFF
--- a/auto_cpufreq/modules/system_info.py
+++ b/auto_cpufreq/modules/system_info.py
@@ -144,21 +144,23 @@ class SystemInfo:
             return None
 
     @staticmethod
-    def current_epp() -> str | None:
+    def current_epp(is_ac_plugged: bool) -> str | None:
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Hi, 

Could you please be more elaborate in what exactly does this PR fix? As I couldn't replicate what changes with `--monitor` with or without config file with what's currently on `master` branch.

PhoenixCausesOof:
Charger is plugged (wrong):
Screenshot From 2025-06-23 13-38-31

Charger is unplugged (correct):
Screenshot From 2025-06-23 13-39-52

The state of EPP and EPB aren't accurate (nor matches with the GUI's data) when the device is connected to AC / being charged. This fixes that. (I apologize for explaining confusingly).

AdnanHodzic:
Thank you for clarification, LGTM. 

PR #858, opened 2025-06-23
fix nix flake

Svenum:
On the latest NixOS unstable version the `requests` override is broken. I updated the flake + removed the unneeded security patches.


shadeyg56:
Thanks for catching this

Issue #604, opened 2023-11-18
[FEATURE] Please allow to turn turbo on and off via override


ufechner7:
- Searched open issues, this one does not seam to exist yet

Please, add the following commands:
```
sudo auto-cpufreq --force=turbo-boost on
sudo auto-cpufreq --force=turbo-boost off
sudo auto-cpufreq --force=turbo-boost auto
``` 
For me, normally I want turbo-boost to be on, but if the battery is below 40% or I go on  a long trip it shall be off.

### System information:
```
Using settings defined in /etc/auto-cpufreq.conf file

-------------------------------------------------------------------------------

Linux distro: Ubuntu 22.04 Jammy Jellyfish
Linux kernel: 6.1.0-1025-oem
Processor: AMD Ryzen 7 7840U w/ Radeon  780M Graphics
Cores: 16
Architecture: x86_64
Driver: acpi-cpufreq

------------------------------ Current CPU stats ------------------------------

CPU max frequency: 3300 MHz
CPU min frequency: 1600 MHz

Core	Usage	Temperature	Frequency
CPU0      1.0%        35 °C      1600 MHz
CPU1      0.0%        35 °C      1600 MHz
CPU2      1.0%        35 °C      1600 MHz
CPU3      1.0%        35 °C      1313 MHz
CPU4      1.0%        35 °C      1600 MHz
CPU5      0.0%        35 °C      1600 MHz
CPU6      2.0%        35 °C      1280 MHz
CPU7      0.0%        35 °C      1600 MHz
CPU8      0.0%        35 °C      1600 MHz
CPU9      0.0%        35 °C      1600 MHz
CPU10      4.0%        35 °C      2152 MHz
CPU11      0.0%        35 °C      1600 MHz
CPU12      0.0%        35 °C      1495 MHz
CPU13      1.0%        35 °C      1600 MHz
CPU14      0.0%        35 °C      1600 MHz
CPU15      1.0%        35 °C      1600 MHz

auto-cpufreq version: 2.0.0 (git: 15c17fc)

Python: 3.10.12
psutil package: 5.9.6
platform package: 1.0.8
click package: 8.1.7
distro package: 1.8.0

Computer type: Notebook
Battery is: discharging

auto-cpufreq system resource consumption:
cpu usage: 0.0 %
memory use: 0.08 %

Total CPU usage: 0.4 %
Total system load: 0.28
Average temp. of all cores: 34.80 °C 

Currently using: ondemand governor
[...]

PR #859, opened 2025-06-26
Allow for overriding the CPU turbo setting via CLI and GUI.

JuanR4140:
As per feature request #604, this pull request implements the ability for the user to override the turbo setting, being able to set it to auto, never, or always. This exists both as a CLI flag and in the GUI (which is only visible if the turbo setting is supported by the CPU.) So far, the CLI flag and GUI setting "work on my machine", as it follows the same principle the performance override uses. Any thoughts, comments, or suggestions are greatly appreciated. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,15 @@ Example of `auto-cpufreq --stats` CLI output
 - [Configuring auto-cpufreq](#configuring-auto-cpufreq)
   - [1: power_helper.py script (Snap package install only)](#1-power_helperpy-script-snap-package-install-only)
   - [2: `--force` governor override](#2---force-governor-override)
-  - [3: auto-cpufreq config file](#3-auto-cpufreq-config-file)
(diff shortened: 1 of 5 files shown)


AdnanHodzic:
This is great! Code LGTM and it works as expected. 

Before I merge this, could you just add a brief/one liner explanation of turbo flag e.g: under `force` [under flags description of README](https://github.com/AdnanHodzic/auto-cpufreq#how-to-run-auto-cpufreq)?

Thanks!

JuanR4140:
Awesome! Happy to hear that it works.

I added info about the turbo flag where it made sense, but if you want me to remove/modify anything, that's totally fine with me!

AdnanHodzic:
This looks great. Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases) as this will be a great new feature.

PR #862, opened 2025-07-11
docs: remove --completions flag README documentation

pranaovs:
Removed in: f9b9fe42c90e176ec4104adcd3eb5e3b4d1823b6

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,9 +479,6 @@ auto-cpufreq should be run with with one of the following options:
 - help
   - Shows all of the above options
 
-- completions=TEXT
-  - To support shell completions (current options are "bash", "zsh", or "fish")
-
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Good catch, indeed this was removed but wasn't done as part of README. 

pranaovs:
May I know why the `--completions` patch was reverted?
Having completions would really ease the use of the cli tool.

AdnanHodzic:
> May I know why the `--completions` patch was reverted? Having completions would really ease the use of the cli tool.

It felt like going too deep into configuring users system and outside of auto-cpufreq scope. Besides this is also didn't exactly work as expected.

If this functionality was implemented as part of auto-cpufreq itself, and didn't need to be a separate flag then I would consider having it back.

Issue #729, opened 2024-07-02
Tray icon


yusifvatan:
Hi. It would be great add the tray icon of application. It would be easy to change power modes

AdnanHodzic:
This is currently only possible with [auto-cpufreq GUI](https://github.com/AdnanHodzic/auto-cpufreq#install---auto-cpufreq-daemon). 

If you or anyone else wants to give it a try in implementing this functionality, please give it a try and [contribute to the project](https://github.com/AdnanHodzic/auto-cpufreq/#code-contribution) and you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

lukasnxyz:
I will have a look at this.

EBendinelli:
Hey there, I decided to give that a go and created a gnome shell extension as it seemed the easiest way for me to get what I wanted. I know it doesn't exactly respond to what's requested here but it might do the trick for some people: https://github.com/EBendinelli/auto-cpufreq-switcher

DISCLAIMER: This is my first gnome-extension so expect issues!

Two notes:
- Password is requested on change, just like with the GUI
- The way I"m getting the current governor is a quick hacky (running `timeout 1s auto-cpufreq --stats` and parsing the output), it would be nice to have a  a `--status`` that simply returns the current governor.

Hope that's useful to someone.

AdnanHodzic:
@EBendinelli this is amazing!

On first look extension is definitively rough around the edges and some things could be improved, but in a nutshell, it works! For 4 years [I had the same idea so it's great to see someone actually did it.](https://github.com/AdnanHodzic/auto-cpufreq/discussions/312)

> In its current form auto-cpufreq is very stable and besides implementing [existing feature requests](https://github.com/AdnanHodzic/auto-cpufreq/issues?q=is%3Aissue+feature+is%3Aopen), next big feature I would need help with is to have it implemented as a [GNOME extension](https://extensions.gnome.org/).

Would you be interested in [creating a PR to work towards merging this with `master` branch?](https://github.com/AdnanHodzic/auto-cpufreq#code-contribution). This would be a great feature and something that would be a highlight of auto-cpufreq 3.0 release.

EBendinelli:
> definitively rough around the edges
I didn't expect anything less after many years without coding haha! But glad to hear it works :)

Happy to have a look into making a proper PR but you'll have to tell me how you envision the integration with the main codebase. 

[...]

PR #863, opened 2025-07-14
Adding gnome extension to README

EBendinelli:
Adding reference to the Gnome Extension in the README with a link to the Gnome extension store as discussed in https://github.com/AdnanHodzic/auto-cpufreq/issues/729

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Example of `auto-cpufreq --stats` CLI output
   - [Supported Devices](#supported-devices)
   - [Battery config](#battery-config)
   - [Ignoring power supplies](#Ignoring-power-supplies)
+  - [🌐 Community Projects](#-community-projects)
 - [Troubleshooting](#troubleshooting)
   - [AUR](#aur)
 - [Discussion](#discussion)
@@ -661,6 +662,12 @@ name4 = example
 xboxctrl = {the xbox controler power supply name}
 
 ```
+
+### 🌐 Community Projects 
+
+If you're using auto‑cpufreq in desktop environments, you may find these community integrations helpful and useful:
+
+- GNOME Extension: [auto-cpufreq-swticher](https://extensions.gnome.org/extension/8320/auto-cpufreq-switcher/) (tray icon + governor switcher in quick settings)
 
 ## Troubleshooting
 


AdnanHodzic:
Great to see this PR! After thoroughly reviewing it, these are my observations and some items that need to be addressed before it can be merged. 

I'm aiming for a new release soon, and I'd love to reference this change as part of it, plus planning on making a new YouTube video talking about all the updates. It's been a while since my last one for the [2.0 release](https://youtu.be/SPGpkZ0AZVU?si=JnhWgqu-CxDcMdcj) and many new features have been implemented.

**Observations and potential new ToDo's:**

**1: Auto-cpufreq Switcher name on GNOME extensions:**

It's a minor thing, but I would call it "auto-cpufreq Profile (governor) Switcher". Or you can leave out governor part.

As part of description on GNOME extensions I would even reference that it does governor switch as explained on: [Configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq?tab=readme-ov-file#configuring-auto-cpufreq) README page.

**2: Enabling extension without having auto-cpufreq daemon installed**

If extension is enabled without auto-cpufreq daemon being installed. Where system is still using GNOME Power Profiles daemon. Extension will show up in GNOME system menu (quick settings) but if toggled it won't do anything. For example I tried switching from balanced to performance.

Extension pop-up will eventually show up (3-5 minutes) saying e.g: "Failed to switch to performance governor".

This could lead to a lot of confusion and I guess same thing will happen if user is using `--live` mode?

It would be ideal if `auto-cpufreq --daemon` is not installed, pop-up would showup and user would be instructed that they need to enable daemon first, e.g: "Please do this by running: `auto-cpufreq --install` or for more info refer to: https://github.com/AdnanHodzic/auto-cpufreq?tab=readme-ov-file#install---auto-cpufreq-daemon"

 **3: "CPU Governor" name in GNOME system menu**
 
 To avoid confusing this with [same setting for "GNOME Power Profiles daemon"](https://x.com/fooctrl/status/1467469508373884933), I would suggest to call it "auto-cpufreq CPU Governor" or something like that. 
 
 **4: If you have auto-cpufreq.conf file, extension will not ignore it and will conflict with it**

If user has [auto-cpufreq.conf file configured](https://github.com/AdnanHodzic/auto-cpufreq?tab=readme-ov-file#4-auto-cpufreq-config-file), it would be ideal if extension would show in pop-up "auto-cpufreq config file detected, please remove otherwise it's conflicting with this auto-cpufreq Profile switcher extension" 

I had a config file configured, and extension still tried to switch governors, but of course these 2 were conflicting with each other.

**5: It takes too long for extension to change governor profiles**

Currently switching governor profiles is taking too long. 

For example, I used `auto-cpufreq --monitor` to track changes and I would switch from Balanced to Performance, I see right command will be triggered and after I enter my root password do this action ... initially nothing happens. 

If I were a new user, I wouldn't have been as patient, would think the extension just doesn't work, and I would delete it.

Eventually a pop-up will say that it successfully changed the governor, and governor will be changed but this will take at least 3-5 minutes in my case and I think this is the biggest problem.

I also tried rebooting, but same behavior persisted. 

Hope this feedback assists in refining this extension :)

AdnanHodzic:
@EBendinelli do you have any updates regarding some of the things I mentioned here? As I'm planning to create a new auto-cpufreq release soon, and would love to also have this PR/GNOME extension mentioned. 

Thanks!

Issue #794, opened 2024-11-18
Requesting for battery threshold management for Asus vivobook S14 S433EA


mantisdc:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? 
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)?
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)?
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)?

### Error output:
```text
Add/paste error output in case of failed installation or other failing component
```
---

### System information:

Add/paste output of:

```
auto-cpufreq --debug
```

I Have an Asus vivobook S14 S433EA and recenty switched to popos and the main issue i had is with battery threshold management(Every other functionality is working totally fine ) as the auto-cpufreq is still not available for this device , i am sharing information about the device driver modules you have asked for but as i am new to linux , i really don't know how to check if the proper acpi modules is installed or not on my linux distro(popos).

here are the info i got from lsmod command:-

Module                  Size  Used by
ccm                    20480  3
rfcomm                102400  4
cmac                   12288  2
algif_hash             12288  1
algif_skcipher         16384  1
af_alg                 32768  6 algif_hash,algif_skcipher
snd_seq_dummy          12288  0
snd_hrtimer            12288  1
zstd                   12288  8
bnep                   32768  2
zram                   36864  2
snd_hda_codec_hdmi     98304  1
snd_sof_pci_intel_tgl    12288  0
snd_sof_intel_hda_common   221184  1 snd_sof_pci_intel_tgl
soundwire_intel        73728  1 snd_sof_intel_hda_common
snd_sof_intel_hda_mlink    45056  2 soundwire_intel,snd_sof_intel_hda_common
soundwire_cadence      40960  1 soundwire_intel
snd_sof_intel_hda      24576  1 snd_sof_intel_hda_common
snd_sof_pci            24576  2 snd_sof_intel_hda_common,snd_sof_pci_intel_tgl
[...]

PR #875, opened 2025-09-04
ASUS laptop to battery thresholds.

PurpleWazard:
i had this fix for a while, its was ment for #794. i can not test this though but ill submit it anyway.


AdnanHodzic:
LGTM, as always thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #882, opened 2025-09-20
fix for asus_expertBook_charging_threshold

sgimres:
My Asus Expertbook laptop use charge_control_end_threshold path instead of charge_stop_threshold
i added charge_control_end_threshold path as a fallback path
and i also add that path to system_info file.
current implementation works with my laptop
Please let me know if any changes are needed.

### DIFF
--- a/auto_cpufreq/battery_scripts/asus.py
+++ b/auto_cpufreq/battery_scripts/asus.py
@@ -5,35 +5,81 @@ from subprocess import check_output
 from auto_cpufreq.config.config import config
 from auto_cpufreq.globals import POWER_SUPPLY_DIR
 
+
 def set_battery(value, mode, bat):
     path = f"{POWER_SUPPLY_DIR}{bat}/charge_{mode}_threshold"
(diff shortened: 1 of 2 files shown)


AdnanHodzic:
@PurpleWazard since you implemented baterry threshold feature, could you please give it a ✅️ from your side before I merge it?

Thanks!

auto_cpufreq/battery_scripts/asus.py
PurpleWazard, line 1: this file looks fine. most of the diff is formatting. please try not to format whole files, unless necessary. it makes the diff really messy.

PurpleWazard:
@AdnanHodzic @sgimres i left notes in the review.  i think that change made to `system_info.py` should be made its own PR to be reviewed.  its out of scope of just battery thresholds.

sgimres:
My bad, my editor uses ruff to auto format python files.
And as per your advice, I split this PR into two separate PRs.
#883 
#884

Issue #886, opened 2025-09-29
Incorect python packages names on tumbleweed (installer needs update)


Mozartwa123:
In the installation script there is a line for opensuse tumbleweed
zypper install -y python3 python3-pip python311-setuptools python3-devel gcc dmidecode gobject-introspection-devel python3-cairo-devel gtk3 gtk3-devel

But most python packages on opensuse tumbleweed requires precise python version, for example python313, instead of python3. So i searched for all packages like in line above and no all appeared (by the way instead of "pakiet" it should be package, i use polish version of tumbleweed, this is why this strange word appeared):

mozartwa123@localhost:~> zypper search -x python3 python3-pip python311-setuptools python3-devel gcc dmidecode gobject-introspection-devel python3-cairo-devel gtk3 gtk3-devel

S  | Name                        | Summary                                                          | Type
---+-----------------------------+------------------------------------------------------------------+-------
i  | dmidecode                   | DMI table decoder                                                | pakiet
i+ | gcc                         | The system GNU C Compiler                                        | pakiet
i+ | gobject-introspection-devel | GObject Introspection Development Files                          | pakiet
i+ | gtk3-devel                  | Development files for the GTK+ toolkit library v3                | pakiet
i+ | python311-setuptools        | Download, build, install, upgrade, and uninstall Python packages | pakiet

Maybe packages name should be named like belove (unfortunately i did not find a correct substitution for gtk3)?

mozartwa123@localhost:~> zypper search -x python313 python313-pip python311-setuptools python313-devel gcc dmidecode gobject-introspection-devel python311-cairo-devel gtk3 gtk3-devel

S  | Name                        | Summary                                                           | Type
---+-----------------------------+-------------------------------------------------------------------+-------
i  | dmidecode                   | DMI table decoder                                                 | pakiet
i+ | gcc                         | The system GNU C Compiler                                         | pakiet
i+ | gobject-introspection-devel | GObject Introspection Development Files                           | pakiet
i+ | gtk3-devel                  | Development files for the GTK+ toolkit library v3                 | pakiet
i+ | python311-setuptools        | Download, build, install, upgrade, and uninstall Python packages  | pakiet
i  | python313                   | Python 3 Interpreter                                              | pakiet
i  | python313-devel             | Include Files and Libraries Mandatory for Building Python Modules | pakiet
i  | python313-pip               | A Python package management system 

Of course, zypper install python3 could cause installation python313, but what if some functionality needs python311 package?

[...]

PR #808, opened 2025-01-15
Update auto-cpufreq-installer to update Suse support

CaffineAddic:
Changed python38 to python3 to solve an install error and also python3-setuptools were causing a conflict using python311-setuptools ended up solving that install.

Solved failed installation on OpenSUSE systems.

PR #887, opened 2025-10-05
Fixed python packages names in installing script for OpenSuse Tumbleweed

Mozartwa123:
I changed the python packages names in the zypper command on opensuse tumbleweed in installer. Of course it is rather cosmetic change, because even python3 package is not proper, zypper automatically choose python313. Please let me know if preffered version is 3.11 (for me it works with 3.13). Sorry for not creating separate branch, but I am unexperienced a little bit yet. I ran this script once again and all packages names was proper.

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -138,7 +138,7 @@ elif [ -f /etc/os-release ];then
       ;;
       opensuse-tumbleweed)
         detected_distro "OpenSUSE"
-        zypper install -y python3 python3-pip python311-setuptools python3-devel gcc dmidecode gobject-introspection-devel python3-cairo-devel gtk3 gtk3-devel
+        zypper install -y python313 python313-pip python311-setuptools python313-devel gcc dmidecode gobject-introspection-devel python313-pycairo-devel libgtk-3-0 gtk3-devel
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Note to myself: Closes #886

Regarding the changes you made, I don't use OpenSuse, but how this works in other distros is that you have a "meta" package that always points and installs latest python version.

So. e.g `zypper install -y python3` should always install the latest version and by saying `zypper install -y python313` you're explicitly pinning it to Python 3.13. This way in future it'll break for someone when future OpenSuse release doesn't have Python 3.13 but only e.g: Python 3.15

I think the issue was created in #808 where it's explicitly pinning `python311-setuptools` to 3.11. What would happen if you changed all the proposed lines to something like:

```
zypper install -y python3 python3-pip python3-setuptools python3-devel gcc dmidecode gobject-introspection-devel python3-cairo-devel gtk3 gtk3-devel
```

Does that fix your problem?

Mozartwa123:
I think I made a bad think. I did not realize, it was intended (because - it works, it installs the latest stable version of python, but I thought at first it was error) and probably wasted your time. I will close this pull request.

PR #889, opened 2025-10-23
improve installer script

laur89:
- replace tabs w/ spaces
- quote vars
- replase dash syntax with that of bash (e.g. [[ "$a" || "$b" ]]) as opposed to [ $a -o $ b ]
- fail wast where command returns non-zero

### DIFF
--- a/auto-cpufreq-installer
+++ b/auto-cpufreq-installer
@@ -221,7 +221,7 @@ tool_remove() {
 
   # run uninstall in case of installed daemon
   if [[ -f "$srv_remove" || -f "$AUTO_CPUFREQ_FILE" ]]; then
-    "$AUTO_CPUFREQ_FILE" --remove || exit 1
+    "$AUTO_CPUFREQ_FILE" --remove || echo -e "\nWARNING: Removal of auto-cpufreq binary failed (exit status $?).\n Continuing..." >&2
   else
     echo; echo "Couldn't remove the auto-cpufreq daemon, [$srv_remove] does not exist."
   fi


auto-cpufreq-installer
```diff
@@ -152,116 +156,113 @@ elif [ -f /etc/os-release ];then
...
   remove_file "/usr/local/bin/cpufreqctl.auto-cpufreq"
   remove_file "/var/run/auto-cpufreq.stats"
 
-  remove_file "$APPLICATIONS_PATH/$AUTO_CPUFREQ_GTK_DESKTOP_FILE"
-  update-desktop-database $APPLICATIONS_PATH
+  remove_file "$APPLICATIONS_PATH/$AUTO_CPUFREQ_GTK_DESKTOP_FILE_NAME"
+  update-desktop-database "$APPLICATIONS_PATH"
 
   # remove python virtual environment
-  remove_directory $venv_path
```
laur89, line 241: note this was a bug - lower case `venv_path` was not defined

auto-cpufreq-installer
```diff
@@ -152,116 +156,113 @@ elif [ -f /etc/os-release ];then
...
-    [ -f $1 ] && rm $1
+  remove_file() {
+    [[ -f "$1" ]] && rm "$1"
   }
 
   srv_remove="$AUTO_CPUFREQ_FILE-remove"
 
   # run uninstall in case of installed daemon
-  if [ -f $srv_remove -o -f $AUTO_CPUFREQ_FILE ]; then
-    eval "$AUTO_CPUFREQ_FILE --remove"
```
laur89, line 220: Why was `eval` used here in the first place?

auto-cpufreq-installer
```diff
@@ -152,116 +156,113 @@ elif [ -f /etc/os-release ];then
...
+  if [[ -f "$srv_remove" || -f "$AUTO_CPUFREQ_FILE" ]]; then
+    "$AUTO_CPUFREQ_FILE" --remove || exit 1
```
laur89: Is this ok? Unsure if `$AUTO_CPUFREQ_FILE" --remove` is allowed to exit non-zero in normal operation or not.

auto-cpufreq-installer
AdnanHodzic: Actually it was defined on [line 161 of auto-cpufeq-installer](https://github.com/AdnanHodzic/auto-cpufreq/blob/52194f4db364347626985fed5372eba2b6109cf4/auto-cpufreq-installer#L161):
```
  venv_dir=$VENV_PATH/venv
  mkdir -p "$venv_dir"
  python3 -m venv "$venv_dir"
```

and [`VENV_PATH="/opt/auto-cpufreq`](https://github.com/AdnanHodzic/auto-cpufreq/blob/52194f4db364347626985fed5372eba2b6109cf4/auto-cpufreq-installer#L12)

AdnanHodzic: Also why did you add `update-desktop-database "$APPLICATIONS_PATH"`?

auto-cpufreq-installer
AdnanHodzic: `eval` is used here to ensure that the entire command string, which is constructed from a variable and an argument is correctly parsed and executed by the shell.

auto-cpufreq-installer
AdnanHodzic: Removing eval makes sense in this case. But potential problem is that if $AUTO_CPUFREQ_FILE --remove returns a non-zero exit code when the daemon isn't installed, script will prematurely exit with an error, even though the desired final state (daemon removed) has been achieved.

Maybe it would make sense to log it while allowing the removal command to fail silently (which is often acceptable for removal steps), with something like:

```
if [[ -f "$srv_remove" || -f "$AUTO_CPUFREQ_FILE" ]]; then
    "$AUTO_CPUFREQ_FILE" --remove || echo "\nWARNING: Removal of auto-cpufreq binary failed (exit status $?).\n Continuing ..." >&2
fi
```

AdnanHodzic:
@laur89 also please check the [Copilot suggestions](https://github.com/AdnanHodzic/auto-cpufreq/pull/889#pullrequestreview-3379932276).

auto-cpufreq-installer
laur89: line 161 is in install path, not remove path. 

> and [VENV_PATH="/opt/auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/blob/52194f4db364347626985fed5372eba2b6109cf4/auto-cpufreq-installer#L12)

Bash is not case insensitive.

auto-cpufreq-installer
laur89: Still doesn't make sense.
`eval "/usr/local/bin/auto-cpufreq --remove"` offers us nothing over running it directly.

auto-cpufreq-installer
laur89: > Also why did you add `update-desktop-database "$APPLICATIONS_PATH"`?

I did no such thing. It's highlighted in the diff due to quotation.

auto-cpufreq-installer
```diff
@@ -96,10 +98,10 @@ function tool_install {
 
     # Update PyGObject in pyproject.toml
     # https://github.com/AdnanHodzic/auto-cpufreq/pull/826#issuecomment-2794549837
-    if [ -f ./pyproject.toml ]; then
+    if [[ -f ./pyproject.toml ]]; then
         if grep -q 'PyGObject *= *{[^}]*version *= *"' pyproject.toml; then
-            sed -i "s/\(PyGObject *= *{[^}]*version *= *\"\)[^\"]*\(.*\)/\1$PYGOBJECT_VER\2/" pyproject.toml
-            echo "PyGObject version updated to $PYGOBJECT_VER in pyproject.toml"
+            sed -i 's/\(PyGObject *= *{[^}]*version *= *"\)[^"]*\(.*\)/\1'"$PYGOBJECT_VER"'\2/' pyproject.toml
+            echo "PyGObject version updated to [$PYGOBJECT_VER] in pyproject.toml"
```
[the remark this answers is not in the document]
laur89: Very much intentional indeed.

auto-cpufreq-installer
AdnanHodzic: Sorry to nitpick, but I think this might confuse some users now, as in case auto-cpufreq daemon is installed and you simply run `sudo ./auto-cpufreq-installer --remove` it's all clear when it comes to messaging.

But if daemon is not installed and you run `sudo ./auto-cpufreq-installer --remove`, output can be bit confusing:

```
auto-cpufreq daemon is not installed.


WARNING: Removal of auto-cpufreq binary failed (exit status 1).
 Continuing...

auto-cpufreq tool and all its supporting files successfully removed
 ```
 
 As `Removal of auto-cpufreq binary failed ` ... it didn't fail, it just wasn't there at this point ... and this is indicated as part of previous message, so maybe good to remove:
 
 ```WARNING: Removal of auto-cpufreq binary failed (exit status 1).
 Continuing...
```

after all.

Regarding whole block:
 
 ```
   # run uninstall in case of installed daemon
  if [[ -f "$srv_remove" || -f "$AUTO_CPUFREQ_FILE" ]]; then
    "$AUTO_CPUFREQ_FILE" --remove || echo -e "\nWARNING: Removal of auto-cpufreq binary failed (exit status $?).\n Continuing..." >&2
  else
    echo; echo "Couldn't remove the auto-cpufreq daemon, [$srv_remove] does not exist."
  fi
  ```
 
Basically `else echo; echo "Couldn't remove the auto-cpufreq daemon, [$srv_remove] does not exist."` is confusing and potentially misleading. The `if` check uses two files `($srv_remove` or `$AUTO_CPUFREQ_FILE`), but the `else` message only mentions `$srv_remove` not existing. This message is only accurate if the `if` statement failed because `$srv_remove` didn't exist AND `$AUTO_CPUFREQ_FILE` also didn't exist. The message should ideally reflect the combined condition failure or be more generic.

AdnanHodzic: I [guess copilot ](https://github.com/AdnanHodzic/auto-cpufreq/pull/889#discussion_r2462744632)was kind of referring to same thing.

laur89: Well that was the very reason that place got flagged -- I'm not familiar with the intricacies of `$AUTO_CPUFREQ_FILE` itself and would like some guidance/help there.

AdnanHodzic: `AUTO_CPUFREQ_FILE` is simply a auto-cpufreq binary, placed on that location after `./auto-cpufreq-installer --install`. After which you can run [install](https://github.com/AdnanHodzic/auto-cpufreq#install---auto-cpufreq-daemon) or [live](https://github.com/AdnanHodzic/auto-cpufreq#live), etc ...

When it comes to [`srv_remove`](https://github.com/AdnanHodzic/auto-cpufreq/blob/1466fc5264b3f63136b239944a80602dde8023e7/auto-cpufreq-installer#L216) and what I see on `master` branch, this [change was introduced with](https://github.com/AdnanHodzic/auto-cpufreq/pull/695/files) with #695, a refactor attempt I should've payed more attention during review and probably shouldn't have merged. 

But before this change, [`srv_remove` pointed to `srv_remove="/usr/local/bin/auto-cpufreq-remove"`.](https://github.com/AdnanHodzic/auto-cpufreq/pull/695/files#diff-0e65217541eff5b94bc7a93997ac399c4978216638a128876011c9311346df23L191).

I can't remember what was the case here, but I think this was because these were 2 different scripts (before they were merged into single install script with parameters), one for removal and other for install:

- [`srv_install="/usr/local/bin/auto-cpufreq-install"`](https://github.com/AdnanHodzic/auto-cpufreq/pull/695/files#diff-0e65217541eff5b94bc7a93997ac399c4978216638a128876011c9311346df23L188)
- [`srv_remove="/usr/local/bin/auto-cpufreq-remove"`](https://github.com/AdnanHodzic/auto-cpufreq/pull/695/files#diff-0e65217541eff5b94bc7a93997ac399c4978216638a128876011c9311346df23L191)

So I think this variable can even be removed now, as it doesn't point to anything. After auto-cpufreq is installed with auto-cpufreq-installer it'll only have 2 binaries in `/usr/local/bin/ `

- auto-cpufreq
- auto-cpufreq-gtk

Issue #833, opened 2025-04-08
Pop OS install fails


noncarbonatedclack:
#### Before submitting an issue, it is strongly recommended to use the **[auto-cpufreq-genAI-chatbot](https://foolcontrol.org/?p=4903)** to get an immediate answer to your question.

The chatbot provided a link that gave me an error 403 forbidden.

### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? - Yes, there are similar issues
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)? - No, i do not use snaps
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)? - No, install fails
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)? - No, install fails

### Error output:
```
> sudo auto-cpufreq/auto-cpufreq-installer 

───────────────────────────────────────────────────────────────────────── auto-cpufreq installer ─────────────────────────────────────────────────────────────────────────

Welcome to auto-cpufreq tool installer.

Select a key [I]nstall/[R]emove or press ctrl+c to quit: i

libgirepository-2.0-dev needs to be installed for version bookworm/sid
---- 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package libgirepository-2.0-dev
E: Couldn't find any package by glob 'libgirepository-2.0-dev'

────────────────────────────────────────────────────────────────── Installing necessary Python packages ──────────────────────────────────────────────────────────────────

Requirement already satisfied: pip in /opt/auto-cpufreq/venv/lib/python3.10/site-packages (25.0.1)
Requirement already satisfied: wheel in /opt/auto-cpufreq/venv/lib/python3.10/site-packages (0.45.1)
Collecting PyGObject
  Using cached pygobject-3.52.3.tar.gz (1.2 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
[...]

PR #892, opened 2025-10-26
Fix PyGObject installation on Pop!_OS

Karthiktiwari:
Issue
Installation fails on Pop!_OS 22.04 and similar Ubuntu systems because PyGObject tries to compile against girepository-2.0, which isn't available #833 . These systems only have libgirepository1.0-dev and already include a working python3-gi package.

Changes
- Create venv with --system-site-packages flag to access system Python packages
- Install python3-gi system package in apt dependencies

Tested on Pop!_OS 22.04 LTS.


AdnanHodzic:
Great job @Karthiktiwari, LGTM and also tested on my side and works as intended! Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases). 

Closes #833

PR #895, opened 2025-10-29
feat: allow users to specify battery device in config

cayubweeums:
Prior, there was no way for users encountering incorrect battery detection via `get_battery_path()` had no option to manually set the desired battery. On my Minisforum V3 tablet I have 3 devices listed in `/sys/class/power_supply/`. The battery detection was selecting the device that resulted in the battery status never changing, and the percentage to always show as 0%. These changes allow the user to specify, via the config, a device from `/sys/class/power_supply` they wish to have auto-cpufreq use. This is done by:

1. Adding `battery_device = {some batt}` as a valid setting in the config file
2. Adding an if statement to `get_battery_path()` to check for the config value above and return the user set device's path if it is a valid device

### DIFF
--- a/auto-cpufreq.conf-example
+++ b/auto-cpufreq.conf-example
@@ -51,6 +51,10 @@ turbo = auto
 
 # settings for when using battery power
 [battery]
+# Specify which battery device to use for reading battery information. see available batteries by running: ls /sys/class/power_supply/
+# If not set, auto-cpufreq will automatically detect and use the first battery found
+# battery_device = BAT1
+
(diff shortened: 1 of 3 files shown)


AdnanHodzic:
@PurpleWazard anything to add from your side before I merge these changes?

PurpleWazard:
Looks good to me 👍

AdnanHodzic:
> Looks good to me 👍

Thanks confirming, because it also LGTM from my side.

Thank you for your contribution, @cayubweeums, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #899, opened 2025-11-13
fix: cpuinfo and scaling min/max freqs not being used properly

CasperVM:
related to:
- #866 
- #737

Bit of a copy paste because this is sort of a fix/revert of a previous issue but here we go:

### Background

In some systems, the two CPU frequency values from `cpuinfo_max_freq` and `scaling_max_freq` return different results. This discrepancy causes the application to pick a lower maximum frequency, which leads to incorrect behavior, such as not going back to the proper max frequency in certain cases.

### What this change does

This update makes the application use the value from `scaling_max_freq` OR `cpuinfo_max_freq` (whichever is bigger), to determine the CPU's maximum frequency. By doing so, it fixes the detection issue on affected systems. It seems that Intel and AMD cpu's work in the opposite way regarding these variables?

For sanity's sake, I also applied this in reverse to the min frequency.

### Impact

For users whose systems do not have this discrepancy, this change should have no impact and will continue working as before. 

For users with this discrepancy, this fixes frequency scaling again.

### DIFF
--- a/scripts/cpufreqctl.sh
+++ b/scripts/cpufreqctl.sh
@@ -203,12 +203,12 @@ function set_frequency_max () {
 
 function get_frequency_min_limit () {
   if [ -z $CORE ]; then CORE=0; fi
-  cat $FLROOT/cpu$CORE/cpufreq/cpuinfo_min_freq
+  echo $(awk '{a[NR]=$1} END{if(a[1]<a[2]) print a[1]; else print a[2]}' $FLROOT/cpu$CORE/cpufreq/cpuinfo_min_freq $FLROOT/cpu$CORE/cpufreq/scaling_min_freq)
 }
 
 function get_frequency_max_limit () {
   if [ -z $CORE ]; then CORE=0; fi
-  cat $FLROOT/cpu$CORE/cpufreq/scaling_max_freq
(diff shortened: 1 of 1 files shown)


scripts/cpufreqctl.sh
```diff
@@ -203,12 +203,12 @@ function set_frequency_max () {
 
 function get_frequency_min_limit () {
   if [ -z $CORE ]; then CORE=0; fi
-  cat $FLROOT/cpu$CORE/cpufreq/cpuinfo_min_freq
+  echo $(awk '{a[NR]=$1} END{if(a[1]<a[2]) print a[1]; else print a[2]}' $FLROOT/cpu$CORE/cpufreq/cpuinfo_min_freq $FLROOT/cpu$CORE/cpufreq/scaling_min_freq)
```
AdnanHodzic, line 206: Nice, I very much like the logic behind comparing what is the lower value between: 

- $FLROOT/cpu$CORE/cpufreq/cpuinfo_min_freq
- $FLROOT/cpu$CORE/cpufreq/scaling_min_freq

While cat before only outputted the contents of one file, which may not reflect the real hardware or current frequency settings.

AdnanHodzic:
LGTM! Think you used some very nice logic here and on my side I'm getting much better ranges then before. Only weird thing is that all my core were initially pinned to 400Mhz until I restarted, which I don't quite have the explanation for.

Regardless, great job, thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #902, opened 2025-11-16
Fix variable name for virtual environment removal

developpeur-php:
Fix: wrong variable name venv_path → VENV_PATH in script

Because the environment variable was misspelled (venv_path instead of VENV_PATH), the script never detected the directory and therefore failed to remove the folder. This commit fixes the variable name so the directory is correctly deleted.


AdnanHodzic:
Good catch! This was [also addressed as part](https://github.com/AdnanHodzic/auto-cpufreq/pull/889#discussion_r2454594920) of #889 but unfortunately these changes haven't been merged yet. 

Issue #812, opened 2025-03-06
How to configure permissions to not require pswd for Governor/power mode change?


Heavygrass:
Hi!

Is it possible to set file permissions in Linux (or do anything else) to **avoid** auto-cpufreq **requiring password** for each **Governor/power mode change**?
Its currently asking for (root permissions? I guess) every time I switch Governor/power mode (at least on Cinnamon).
If this is possible, which files one need to change ownership or permissions?

I did a search of issues and couldn't find anything relevant, please let me know if there's.

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted? 
Yes.

Thank you for this great project!

shadeyg56:
The files `auto-cpufreq` edits are in sysfs which is provided by the kernel and they come from the `cpufreq` driver. It is up to the driver to set the owner of the attributes it exposes. You might be able to make a group and combine that with some udev rules or something, but I wouldn't recommend messing with it too much since it could lead to unexpected behavior.

AdnanHodzic:
@Heavygrass you can simply configure [sudo NOPASSWD in sudoers file](https://www.tecmint.com/run-sudo-command-without-password-linux/) not to be asked for password every time you want to run something as root.

Heavygrass:
@AdnanHodzic Thank you for your response!
Is removing root password the only way?

AdnanHodzic:
Who said anything about removing root password? 🤔

Article I shared above instructs you how to not be prompted for password every time you want to run something as root.

Heavygrass:
@shadeyg56 @AdnanHodzic 
Is it possible to point me to say a discussion (or anything else) that could help me to understand the reason(s) for requiring sudo for governor change design decision was made?
Thank you.

AdnanHodzic:
You're asking questions that don't have anything to do with auto-cpufreq and are related to how Linux itself and some of its operations were designed.

If you want to change e.g CPU governor state in Linux, you need sudo/root permissions to perform this action (you can see this in code btw). Hence if you want auto-cpufreq to automatically make these decisions for you, you need to provide it with root access, simple as that.

I can't find exact discussions now (which are somewhere in issues), but that's why auto-cpufreq won't ask you for root for actions where it's not needed. For example, you can run `auto-cpufreq --stats` (without sudo) as it's not making any actions and is just printing information.

ekasprzak:
[...]

PR #903, opened 2025-11-18
Improve pkexec error handling for GUI

developpeur-php:
I had issues with the GUI, so I fixed the code to handle it better for most users.

pkexec errors were being handled by verifying the string returned by the pkexec command.

I noticed the expected string — "Error executing command as another user: Not authorized\n\nThis incident has been reported.\n" — did not match what I had on Arch Linux.

So I checked the pkexec manual to verify the return codes and corrected the code accordingly.

> RETURN VALUE
       Upon successful completion, the return value is the return value of
       PROGRAM. If the calling process is not authorized or an authorization
       could not be obtained through authentication or an error occurred,
       pkexec exits with a return value of 127. If the authorization could not
       be obtained because the user dismissed the authentication dialog, pkexec
       exits with a return value of 126.

before
before

after
after

### DIFF
--- a/auto_cpufreq/gui/app.py
+++ b/auto_cpufreq/gui/app.py
@@ -82,7 +82,7 @@ class ToolWindow(Gtk.Window):
         dialog.destroy()
         if response != Gtk.ResponseType.YES: return
         updater = run(["pkexec", "auto-cpufreq", "--update"], input="y\n", encoding="utf-8", stderr=PIPE)
-        if result.returncode in (126, 127):
+        if updater.returncode in (126, 127):
             error = Gtk.MessageDialog(self, 0, Gtk.MessageType.ERROR, Gtk.ButtonsType.OK, "Error updating")
             error.format_secondary_text("Authorization Failed")
             error.run()


AdnanHodzic:
Nice one @developpeur-php! LGTM, Thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases). 

Btw, @ekasprzak and @Heavygrass do these changes also fix your problem in #812?

Heavygrass:
> Btw, @ekasprzak and @Heavygrass do these changes also fix your problem in #812?

Is there a built for Ubuntu 24.04(.3)?

AdnanHodzic:
> > Btw, @ekasprzak and @Heavygrass do these changes also fix your problem in #812?
> 
> Is there a built for Ubuntu 24.04(.3)?

I was hoping that by now everyone knows that [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq?tab=readme-ov-file#auto-cpufreq-installer) supports all major linux distributions, regardless of their release.

PR #916, opened 2026-01-18
Add monitoring mode to the GUI

cypherpower:
Adds the monitoring mode to the GUI where changes can be previewed before an install.  Matches the layout from the CLI.

PR #917, opened 2026-01-19
Add bluetooth controls to the GUI, and fix bluetooth conf updates for commented lines

cypherpower:
This adds a new radio dialog for in the GUI for setting the bluetooth modes similar to the CLI. I noticed when testing it was not working, the previous impelementation was using a find and replace string to update the main.conf file for bluetooth configurations, which for me on arch linux was just updating a commented line in the configuration file which is the how the file is setup by default.
```bash
# AutoEnable defines option to enable all controllers when they are found.
# This includes adapters present on start as well as adapters that are plugged
# in later on. Defaults to 'true'.
#AutoEnable=false
```

I added more robust parsing to determine current bluetooth settings and updated the replacement mechanism to deal with commented lines in addition to proper location placement under the [Policy] section in the bluetooth config file.

### DIFF
--- a/auto_cpufreq/core.py
+++ b/auto_cpufreq/core.py
@@ -870,6 +870,13 @@ def sysinfo():
     driver = getoutput("cpufreqctl.auto-cpufreq --driver")
     print("Driver: " + driver)
 
+    config_path = config.path if config.has_config() else None
+    if config_path is None:
+        from auto_cpufreq.config.config import find_config_file
+        config_path = find_config_file(None)
+    if os.path.isfile(config_path):
+        print(f"\nUsing settings defined in {config_path}")
+
     # get usage and freq info of cpus
     usage_per_cpu = psutil.cpu_percent(interval=1, percpu=True)
     # psutil current freq not used, gives wrong values with offline cpu's
@@ -936,9 +943,12 @@ def sysinfo():
 
     if offline_cpus: print(f"\nDisabled CPUs: {','.join(offline_cpus)}")
 
-    # print current fan speed
+    # print current fan speed (only if > 0)
     current_fans = list(psutil.sensors_fans())
-    for current_fan in current_fans: print("\nCPU fan speed:", psutil.sensors_fans()[current_fan][0].current, "RPM")
+    for current_fan in current_fans:
+        fan_speed = psutil.sensors_fans()[current_fan][0].current
+        if fan_speed:
+            print(f"\nCPU fan speed: {fan_speed} RPM")
 
 def read_stats():
     if os.path.isfile(auto_cpufreq_stats_path): call(["tail", "-n 50", "-f", str(auto_cpufreq_stats_path)], stderr=DEVNULL)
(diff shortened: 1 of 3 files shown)


AdnanHodzic:
This is amazing and great to see auto-cpufreq GUI is getting so much needed love.

However, now with "Bluetooth on boot" option being added, the interface feels bit cluttered:
auto-cpufreq GUI cluttered

Might be a good idea to replace these (this area) with a button just like you did with Monitor mode in #916, and name the button as "Advanced & manual settings ⚙️"? I'm open to suggestions and looking forward to hear what you think about this.

If you agree to doing this, if it's too much to make changes as part of this PR, I could merge current changes, and then you create a separate PR where the button is implemented and interface is cleared up?

Btw, another thing that I noticed is missing from "informational" part of GUI, if auto-cpufreq.conf file is used, it will not be noted it's used anywhere in GUI like it is in CLI.

CLI clearly indicating it's used:
auto-cpufreq --stats using config file

Yet nothing on GUI even if file is used:
auto-cpufreq GUI

More things that are not shown/reflected in GUI but are displayed in CLI and would also be great in GUI and could affect interface layout:

* Discharge level in watts (W) - when on battery power
* Battery status: charging - when charging
* AC plugged: yes/no
* Battery percentage
* Charging start threshold
* Charging stop threshold

cypherpower:
> Might be a good idea to replace these (this area) with a button just like you did with Monitor mode in #916, and name the button as "Advanced & manual settings ⚙️"? I'm open to suggestions and looking forward to hear what you think about this.
> 
I agree it is a bit cluttered.
Added a line on the GUI to show the config file in use and moved bluetooth options to an advanced settings menu
image

> More things that are not shown/reflected in GUI but are displayed in CLI and would also be great in GUI and could affect interface layout:
> 
> * Discharge level in watts (W) - when on battery power
> * Battery status: charging - when charging
> * AC plugged: yes/no
> * Battery percentage
> * Charging start threshold
> * Charging stop threshold

Also added these details to match the CLI.

cypherpower:
Still don't really like how the GUI looks but plan to work on this later once I get all of the features added. Please let me know if you think anything else could use a quick visual update though.  I wanted to get your thoughts on a potential upgrade to GTK4?  It would make it easier for me to work on the UI layout, not sure if we need this to remain in GTK3 though for compatibility reasons? From the quick research I did it looks like all of the major linux distributions support GTK4 now including mint which looked like it was a holdout for a bit?  So it doesn't seem like there should be problems there, you may have more knowledge in this regard though.

AdnanHodzic:
> Still don't really like how the GUI looks but plan to work on this later once I get all of the features added.

Solid plan, even with interface bit cluttered it looks much better with all the information showing up. I would maybe even move "Governor Override" and "CPU Turbo Override" under "Advanced Settings". So initial screen would only show information and could look cleaner that way.

>  I wanted to get your thoughts on a potential upgrade to GTK4? It would make it easier for me to work on the UI layout, not sure if we need this to remain in GTK3 though for compatibility reasons? From the quick research I did it looks like all of the major linux distributions support GTK4 now including mint which looked like it was a holdout for a bit? So it doesn't seem like there should be problems there, you may have more knowledge in this regard though.

I think this would be a good idea. Every distro has been moving to GTK4 for some time now, with upcoming release of Ubuntu 26.04 in April which is making big shift to GTK4, and then think rest of distros like Mint will have to follow along since they are based on it anyways. 

Regardless, great changes in this PR, as always thank you for your contribution and will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases).

PR #926, opened 2026-02-16
Fix power consumption calculation for systems missing power_now file

shadeyg56:
Wattage values were being incorrectly calculated if the `power_now` file was not found in the battery directory. Before it was just reading from the path `current_now` if `power_now` wasn't found, which just returned the current in Amps.

This should fix it so if `power_now` is not found, it will calculate the power draw using current and voltage instead, if possible.


AdnanHodzic:
Great, thank you!

PR #928, opened 2026-02-25
Change auto-cpufreq from programs to services

CasperGamingOne:
for me it only works when it is exactly like it is in nixpkgs, idk if only for me or not; or anything changed, but if it did change i think this should be updated. Hope i am not stupid 🫣

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,9 +226,9 @@ Then you can enable the program in your `configuration.nix` file:
 
 {inputs, pkgs, ...}: {
     # ---Snip---
-    programs.auto-cpufreq.enable = true;
+    services.auto-cpufreq.enable = true;
     # optionally, you can configure your auto-cpufreq settings, if you have any
-    programs.auto-cpufreq.settings = {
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
Since it concerns NixOS will leave this one for you @shadeyg56

shadeyg56:
The correct setting if you are using the flake is `programs.auto-cpufreq`. This is different from the setting declared in nixpkgs which is `services.auto-cpufreq`. 

The [README](lhttps://github.com/AdnanHodzic/auto-cpufreq?tab=readme-ov-file#nixos) contains different instructions for installing via the flake or from the nixpkgs repo, so the current version has the correct instructions.

If you are installing via the flake and having issues with `programs.auto-cpufreq` not being recognized, its most likely that something is wrong with your nix config. If that's the case and you'd like some help feel free to join the [Discord](https://discord.com/invite/Sjauxtj6kH) and I can help you with that!

PR #929, opened 2026-02-25
fix: add missing dependencies to the nix package

nicobonada:
kmod (needed for lsmod) and dmidecode


nicobonada:
Hi, I started getting an error due to lsmod being added in pr #900. Adding kmod to the list of packages fixes that.

dmidecode is needed for `auto-cpufreq --debug` to display the Computer Type

AdnanHodzic:
@shadeyg56 will leave this one for you :)

shadeyg56:
I made a small change to move the deps into the same list instead of using list concatenation.

Thanks for your contribution!

Issue #927, opened 2026-02-24
EPP write fails with "Device or resource busy" on Intel Core Ultra 7 155H (Arch)


Z-Mackintosh:
## Summary
On my machine, `auto-cpufreq` works for governor switching, but writing EPP fails consistently.

`cpufreqctl.auto-cpufreq --epp --set=performance` returns:

```
/usr/bin/cpufreqctl.auto-cpufreq: line 99: echo: write error: Device or resource busy
```

The error repeats once per logical CPU (22 times on this machine).

## Environment
- Distro: EndeavourOS (Arch-based, rolling)
- Kernel: `6.18.9-arch1-2`
- CPU: `Intel(R) Core(TM) Ultra 7 155H`
- Driver: `intel_pstate` (`active`)
- auto-cpufreq: installed as systemd service (`auto-cpufreq.service`)

## What works
- `sudo cpufreqctl.auto-cpufreq --governor --set=performance` works
- `auto-cpufreq.service` runs and applies other settings
- `power-profiles-daemon` is masked/inactive

## What fails
- `sudo cpufreqctl.auto-cpufreq --epp --set=performance`
- same failure even after changing platform profile from `low-power` to `balanced`

## Reproduction
```bash
sudo cpufreqctl.auto-cpufreq --governor --set=performance
sudo cpufreqctl.auto-cpufreq --epp --set=performance
```

Also tested:

```bash
cat /sys/firmware/acpi/platform_profile
cat /sys/firmware/acpi/platform_profile_choices
echo balanced | sudo tee /sys/firmware/acpi/platform_profile
sudo cpufreqctl.auto-cpufreq --epp --set=performance
```

## Additional observations
- EPP choices are present:
  - `/sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences`
  - values: `default performance balance_performance balance_power power`
- Current EPP across policies remains `default`
- Error occurs per-core/per-policy write attempt

## Expected behavior
If EPP cannot be changed on this hardware/firmware state, `auto-cpufreq` should gracefully skip EPP writes (or fallback) without flooding logs with repeated errors.

## Question
Is this a known limitation for newer Intel Core Ultra + current kernel/firmware combinations, and should EPP be auto-skipped in this case?

AdnanHodzic:
Changes related to wrong EPP and EBP fallback profiles were merged as part of #925 

But this issue seems to be **identical to what was reported in #751** and similar things were reported in #724

[...]

Issue #741, opened 2024-07-17
Spamming of 'line 105: echo: write error: Device or resource busy' in journal (once again)


arcenik:
### Have you tried?

- [Searching through existing/closed issues](https://github.com/AdnanHodzic/auto-cpufreq/issues) to see if your bug has already been already submitted?  yes
- If installation failed with [auto-cpufreq-installer](https://github.com/AdnanHodzic/auto-cpufreq/#auto-cpufreq-installer),have you tried installing auto-cpufreq using [Snap package](https://github.com/AdnanHodzic/auto-cpufreq/#snap-store)? no
- Have you tried [configuring auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq/#configuring-auto-cpufreq)? yes
- Have you tried suggestions in [troubleshooting section](https://github.com/AdnanHodzic/auto-cpufreq#troubleshooting)? no yet

### Error output:

Yet another issue with the line 105 echo write error.
The system is a Legion 7 laptop with AMD Ryzen 7 5800H with Radeon Graphics CPU.
This affect only the performance mode.

```text
# systemctl status auto-cpufreq.service
● auto-cpufreq.service - auto-cpufreq - Automatic CPU speed & power optimizer for Linux
     Loaded: loaded (/usr/lib/systemd/system/auto-cpufreq.service; enabled; preset: disabled)
     Active: active (running) since Wed 2024-07-17 10:40:51 CEST; 27min ago
 Invocation: e27bc93997ca449a97868688b47e26a8
   Main PID: 6187 (auto-cpufreq)
      Tasks: 2 (limit: 76915)
     Memory: 23.9M (peak: 27M)
        CPU: 24.727s
     CGroup: /system.slice/auto-cpufreq.service
             └─6187 /usr/bin/python /usr/bin/auto-cpufreq --daemon

jui 17 11:08:20 alpha.arcenik.lan auto-cpufreq[43239]: /usr/bin/cpufreqctl.auto-cpufreq: line 105: echo: write error: Device or resource busy
jui 17 11:08:20 alpha.arcenik.lan auto-cpufreq[43239]: /usr/bin/cpufreqctl.auto-cpufreq: line 105: echo: write error: Device or resource busy
jui 17 11:08:20 alpha.arcenik.lan auto-cpufreq[43239]: /usr/bin/cpufreqctl.auto-cpufreq: line 105: echo: write error: Device or resource busy
[...]

Issue #891, opened 2025-10-24
powersave is on, but governer locked on performance - logs: NotImplementedError: can't find current frequency file


ElijahLynn:
### Summary
powersave is on, but governer locked on performance - logs: NotImplementedError: can't find current frequency file

<img width="3636" height="2085" alt="Image" src="https://github.com/user-attachments/assets/fb476a9e-6eb8-4a7d-afff-accf8deab3ce" />

Clicking "about" in auto cpufreq menu, results in nothing happening.

### Error output:
```text
systemctl status auto-cpufreq.service
● auto-cpufreq.service - auto-cpufreq - Automatic CPU speed & power optimizer for Linux
     Loaded: loaded (/usr/lib/systemd/system/auto-cpufreq.service; enabled; preset: disabled)
     Active: active (running) since Thu 2025-10-23 20:55:42 PDT; 13h ago
 Invocation: edd183be68214c91a6fca5f31108f0a9
   Main PID: 796 (auto-cpufreq)
      Tasks: 2 (limit: 37925)
     Memory: 82.6M (peak: 88M)
        CPU: 1min 11.999s
     CGroup: /system.slice/auto-cpufreq.service
             └─796 /usr/bin/python /usr/bin/auto-cpufreq --daemon

Oct 23 21:29:20 gaia-two auto-cpufreq[796]:   File "/usr/lib/python3.13/site-packages/auto_cpufreq/modules/system_info.py", line 179, in avg_temp
Oct 23 21:29:20 gaia-two auto-cpufreq[796]:     temps: List[float] = [i.temperature for i in SystemInfo.get_cpu_info()]
Oct 23 21:29:20 gaia-two auto-cpufreq[796]:                                                  ~~~~~~~~~~~~~~~~~~~~~~~^^
Oct 23 21:29:20 gaia-two auto-cpufreq[796]:   File "/usr/lib/python3.13/site-packages/auto_cpufreq/modules/system_info.py", line 104, in get_cpu_info
Oct 23 21:29:20 gaia-two auto-cpufreq[796]:     cpu_freqs = psutil.cpu_freq(percpu=True)
Oct 23 21:29:20 gaia-two auto-cpufreq[796]:   File "/usr/lib/python3.13/site-packages/psutil/__init__.py", line 1913, in cpu_freq
Oct 23 21:29:20 gaia-two auto-cpufreq[796]:     ret = _psplatform.cpu_freq()
[...]

Issue #842, opened 2025-05-04
[Bug] /usr/bin/cpufreqctl.auto-cpufreq: line 99: echo: write error: Device or resource busy


itzayush18:
> I'm on Arch Linux on a Lenovo IdeaPad Slim 5 with an Intel Core Ultra 5 125H CPU, and I still get this error when in charging mode.
> 
> Exact error:
> ```
> /usr/bin/cpufreqctl.auto-cpufreq: line 99: echo: write error: Device or resource busy
> ``` 

 _Originally posted by @itzayush69 in [#751](https://github.com/AdnanHodzic/auto-cpufreq/issues/751#issuecomment-2849155666)_

![Image](https://github.com/user-attachments/assets/d599dac4-d457-40b3-85ca-7f5e0f791bfb)

**Note:** This error only occurs when the laptop is connected to AC power (plugged in).

Below is the config (```/etc/auto-cpufreq.conf```):
_I haven't changed anything, just added `ideapad_laptop_conservation_mode = true`_
```
# settings for when connected to a power source
[charger]
# see available governors by running: cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors
# preferred governor
governor = performance

# EPP: see available preferences by running: cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
energy_performance_preference = performance

# EPB (Energy Performance Bias) for the intel_pstate driver
# see conversion info: https://www.kernel.org/doc/html/latest/admin-guide/pm/intel_epb.html
# available EPB options include a numeric value between 0-15
# (where 0 = maximum performance and 15 = maximum power saving),
# or one of the following strings:
# performance (0), balance_performance (4), default (6), balance_power (8), or power (15)
# if the parameter is missing in the config and the hardware supports this setting, the default value will be used
# the default value is `balance_performance` (for charger)
# energy_perf_bias = balance_performance
energy_perf_bias = performance

# Platform Profiles
# https://www.kernel.org/doc/html/latest/userspace-api/sysfs-platform_profile.html
# See available options by running:
# cat /sys/firmware/acpi/platform_profile_choices
# platform_profile = performance

# minimum cpu frequency (in kHz)
# example: for 800 MHz = 800000 kHz --> scaling_min_freq = 800000
# see conversion info: https://www.rapidtables.com/convert/frequency/mhz-to-hz.html
# to use this feature, uncomment the following line and set the value accordingly
[...]

PR #958, opened 2026-08-17
Fix cpufreqctl reporting success after write failures

lucasleocs:
## Summary

Fix `cpufreqctl` reporting successful execution when a sysfs write was rejected by the kernel.

Currently, a write can fail with a real kernel error while`cpufreqctl` still exits with status `0`. This prevents callers from distinguishing a successfully applied setting from a failed or partially applied one.

This PR preserves the existing best-effort behavior for multi-CPU operations, but makes the final exit status correctly report whether any attempted write failed.

## Problem

`cpufreqctl` currently does not reliably propagate sysfs write failures.

For example, a write rejected by the kernel can produce:

```text
echo: write error: Device or resource busy
return code: 0
```

The command therefore reports success even though the requested operation was not successfully completed.

The same incorrect success status can be reproduced independently of CPU hardware by using `/dev/full` as the write target, confirming that the issue is in `cpufreqctl`'s status handling rather than a specific CPU driver.

There are two cases where the failure status is lost:

1. Multi-CPU setters continue iterating after a failed write, so the status of a later successful operation can hide an earlier failure.
2. The unconditional `exit 0` at the end of the script masks failures from direct write operations.

For multi-CPU operations this can also result in a partially applied configuration being reported as fully successful.

Besides being misleading when `cpufreqctl` is used directly, this prevents callers from reliably detecting and handling write failures.

## Changes

- Track failed writes during multi-CPU operations while continuing to attempt the remaining CPUs.
- Apply the same handling to the multi-CPU EPB path.
- Preserve the command status of direct write operations.
- Return non-zero if at least one attempted write failed.
- Preserve the existing exit behavior of read-only commands.
- Preserve the original kernel error on stderr.

Existing auto-cpufreq callers that do not inspect the return code continue running as before. Callers that do inspect it can now correctly detect failed writes.

## Validation

I tested the change with 9 synthetic cases covering:

* successful multi-CPU EPP writes
* partial multi-CPU failures while continuing to later CPUs
* single-core EPP failures
* governor write failures
* direct boost write failures
* multi-CPU EPB failures
* successful normal commands
* preservation of existing read-command exit behavior
* `--off` write failures

All 9 synthetic cases passed.

I also tested the patch on an Intel i3-1315U with `intel_pstate` active.

Before the patch:

```text
rejected EPP write:
Device or resource busy
return code: 0
```

With the patch:

```text
rejected single-core EPP write:
return code: 1

rejected multi-CPU EPP writes:
8 x Device or resource busy
return code: 1

compatible EPP=performance write:
return code: 0
```

The original governor/EPP state was restored after testing and the auto-cpufreq service was restarted successfully.

## Relation to #927

This does not attempt to fix the underlying `EBUSY` behavior reported in #927.

The specific Core Ultra behavior did not reproduce on my system. This patch fixes an independent issue discovered during that investigation: when the kernel rejects a sysfs write, `cpufreqctl` should not report the operation as successful.

Correct exit status propagation may also make failures such as #927 easier for callers to detect and handle.

Related: #741, #927

### DIFF
--- a/scripts/cpufreqctl.sh
+++ b/scripts/cpufreqctl.sh
@@ -6,6 +6,7 @@ FLROOT=/sys/devices/system/cpu
 FWROOT=/sys/firmware
 DRIVER=auto
 VERBOSE=0
+WRITE_ERROR=0
 
 ## parse special options
 for i in "$@"; do
@@ -96,7 +97,11 @@ function driver () {
 }
 
 function write_value () {
-  if [ -w $FLNM ]; then echo $VALUE > $FLNM; fi
+  if [ -w $FLNM ]; then
+    if ! echo $VALUE > $FLNM; then
+      WRITE_ERROR=1
+    fi
+  fi
 }
 
 function set_driver () {
@@ -285,7 +290,11 @@ function set_energy_performance_bias () {
     i=0
     while [ $i -ne $cpucount ]; do
       FLNM="$FLROOT/cpu"$i"/power/energy_perf_bias"
-      if [ -w $FLNM ]; then echo $EPB_VALUE > $FLNM; fi
+      if [ -w $FLNM ]; then
+        if ! echo $EPB_VALUE > $FLNM; then
+          WRITE_ERROR=1
+        fi
+      fi
       i=`expr $i + 1`
     done
   else echo $EPB_VALUE > $FLROOT/cpu$CORE/power/energy_perf_bias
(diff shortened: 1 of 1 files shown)


AdnanHodzic:
This is a great one to tackle as it's a long standing issue and this problem has been mentioned in 6 different issues: #927, #891, #842, #750, #741, #602. I'll update each one of them and let's see if these and/or other numerous changes made in meantime addressed it.

Amazing stuff, as always thank you for your contribution, you will be credited for your work as part of [future release](https://github.com/AdnanHodzic/auto-cpufreq/releases)

PR #976, opened 2026-09-10
fix: report HWP Dynamic Boost in debug output

lucasleocs:
## Summary

Report the current Intel HWP Dynamic Boost state in `--debug` when the kernel exposes the control.

This reuses the existing read-only `get_hwp_dynamic_boost()` helper and does not change power-management policy.

## Behavior

- `True` → `HWP Dynamic Boost: On`
- `False` → `HWP Dynamic Boost: Off`
- unavailable or unsupported → no additional line

This makes `--debug` more useful when investigating Intel P-State power behavior while preserving the existing output on systems without HWP Dynamic Boost support.


AdnanHodzic:

